PR #14, opened 2015-04-16
initial draft of documentation.

chandraprakash:


src/connection_manager.rs
```diff
@@ -101,10 +111,26 @@ impl ConnectionManager {
         Ok(vec![Endpoint::Tcp(local_addr)])
     }
 
+    /// This method tries to connect (bootstrap to exisiting network) to the default or provided
+    /// list of bootstrap nodes. If the bootstrap list is `Some`, the method will try to connect to
+    /// all of the endpoints specified in `bootstrap_list`. It will return once connection with any of the
```
dirvine: is this correct -> all of the endpoints in the bootstrap list ?

chandraprakash: I have not mentioned sequentially as it may change.

chandraprakash: It will try the list until it gets one successful connection.

dirvine: OK merging now, sounds reasonable, thx for this

PR #16, opened 2015-04-16
Adds generic transport for ConnectionManager

inetic:
The ConnectionManager no longer cares what protocol is used.


dirvine:
Can you please look at the conflicts.

PR #20, opened 2015-04-17
Adds PartialOrd and Ord interfaces for transport::Endpoint

inetic:


dirvine:
Nice one @inetic Just a few more tiny % for coverage passing as well. Great work man thanks. Soon we will have utp and NAT traversal, then we will be flying.

inetic:
Cheers @dirvine, I'll add more tests for sure. I'm a bit puzzled though by the fact that the check failed while coverage had increased, is it because the increase wasn't high enough?

dirvine:
Yes at the moment our settings are 90% is pass and any reduction > 1% is a fail. I hope to make it any reduction at all is a fail.

PR #23, opened 2015-04-20
beacon example

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/23)

### DIFF
--- a/examples/beacon_server.rs
+++ b/examples/beacon_server.rs
@@ -19,15 +19,16 @@
 extern crate crust;
 
 use std::net::UdpSocket;
+use std::net::TcpListener;
 
 fn main() {
-  let normal_socket = match UdpSocket::bind("::0:0") {
-    Ok(socket) => socket,
-    Err(e) => panic!("Couldn't bind socket: {}", e)
+  let tcp_listener = match TcpListener::bind("0.0.0.0:0") {
+    Ok(listener) => listener,
+    Err(e) => panic!("Couldn't bind to TCP socket: {}", e)
   };
 
   // blocking call on listen_for_broadcast
-  match normal_socket.local_addr() {
+  match tcp_listener.local_addr() {
     Ok(local_addr) => crust::beacon::listen_for_broadcast(local_addr),
     Err(e) => panic!("No local address to start listening on: {}", e)
   };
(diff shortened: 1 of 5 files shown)


examples/beacon_server.rs
```diff
@@ -0,0 +1,38 @@
...
+
+fn main() {
+  spawn(move || {
+    let normal_socket = match UdpSocket::bind("::0:0") {
+      Ok(socket) => socket,
+      Err(e) => panic!("Couldn't bind socket: {}", e)
+    };
+
+    match normal_socket.local_addr() {
+      Ok(local_addr) => crust::beacon::listen_for_broadcast(local_addr),
```
benjaminbollen: I don't seem to reach this point, neither do I get panics;  are panics in a spawned off thread invisible on terminal?

benjaminbollen:
works on one machine.  will test now on several machines, cross platform with multiple nodes

rossmuir:
As per yesterdays discussion - should `#deny(missing_docs)` be added to this PR?

src/beacon.rs
```diff
@@ -16,6 +16,8 @@
 // See the Licences for the specific language governing permissions and limitations relating to
 // use of the MaidSafe Software.
 
+#![deny(missing_docs)]
```
rossmuir: Sorry @benjaminbollen my bad - can we switch this on in the `lib.rs` file instead

examples/beacon_server.rs
```diff
@@ -0,0 +1,34 @@
...
+//
+// See the Licences for the specific language governing permissions and limitations relating to
+// use of the MaidSafe Software.
+
+extern crate crust;
+
+use std::net::UdpSocket;
+
+fn main() {
+  let normal_socket = match UdpSocket::bind("::0:0") {
```
chandraprakash: More realistic example will be to use supported protocol. So using tcp here will be handy.  
Later on bootstrap integration can also be used here too.

dirvine: Agreed and IPv4 please! IPv6 can be later iteration

benjaminbollen: Alright, seen the comment.  I'll address it!

src/beacon.rs
benjaminbollen: I will make a new pull request to document instead

chandraprakash:
good to go ..

PR #32, opened 2015-04-21
WIP

BSMaidSafe:
WIP.

[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/32)

### DIFF
--- a/src/bootstrap.rs
+++ b/src/bootstrap.rs
@@ -21,15 +21,11 @@ use cbor;
 use cbor::CborTagEncode;
 use rustc_serialize::{Decodable, Decoder, Encodable, Encoder};
 use sodiumoxide::crypto;
-use std::net;
-use std::net::SocketAddr;
 use transport::Endpoint;
 use std::fs::File;
-use std::io;
 use std::io::prelude::*;
 use std::path;
 use std::env;
-use beacon;
 
 type BootStrapContacts = Vec<Contact>;
 
@@ -49,22 +45,6 @@ macro_rules! convert_to_array {
             Some(arr)
         }
     }};
-}
-
-fn array_to_vec(arr: &[u8]) -> Vec<u8> {
-    let mut vector = Vec::new();
-    for i in arr.iter() {
-        vector.push(*i);
-    }
-    vector
-}
-
-fn vector_as_u8_4_array(vector: Vec<u8>) -> [u8;4] {
-    let mut arr = [0u8;4];
-    for i in (0..4) {
-        arr[i] = vector[i];
-    }
-    arr
 }
 
 // TODO Move Contact to maidsafe_types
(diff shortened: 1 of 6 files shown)


src/transport.rs
```diff
@@ -19,16 +19,55 @@ use std::io;
...
 use std::sync::mpsc;
+use beacon;
+use cbor;
+use cbor::CborTagEncode;
+use rustc_serialize::{Decodable, Decoder, Encodable, Encoder};
 use std::cmp::Ordering;
 
 pub type Bytes = Vec<u8>;
 
+fn array_to_vec(arr: &[u8]) -> Vec<u8> {
```
dirvine: would `arr.iter().collect()` not work here?

src/bootstrap.rs
```diff
@@ -0,0 +1,273 @@
...
+use std::io;
+use std::io::prelude::*;
+use std::path;
+use std::env;
+use beacon;
+
+type BootStrapContacts = Vec<Contact>;
+
+static MAX_LIST_SIZE: usize = 1500;
+
```
dirvine: No unsafe blocks. We should add a utils file in crust and expose this as a public macro
```macro_rules! convert_to_array {
    ($container:ident, $size:expr) => {{
        if $container.len() != $size {
            None
        } else {
            let mut arr = [0u8; $size];
            for element in $container.into_iter().enumerate() {
                arr[element.0] = element.1;
            }
            Some(arr)
        }
    }};
}```

File: src/bootstrap.rs [hunk 3]
```diff
@@ -0,0 +1,273 @@
...
+fn vector_as_u8_4_array(vector: Vec<u8>) -> [u8;4] {
+    let mut arr = [0u8;4];
+    for i in (0..4) {
+        arr[i] = vector[i];
+    }
+    arr
+}
+
+// TODO Move Contact to maidsafe_types
```
dirvine: redundant comment (no need to move)

src/bootstrap.rs
BSMaidSafe: Yes, it was pulled from maidsafe_types.

src/transport.rs
BSMaidSafe: Of course it might :-)

src/bootstrap.rs
dirvine: I asked here @BSMaidSafe lets see if we get some better mechanism here http://stackoverflow.com/questions/29784502/convert-rust-vectors-to-arrays-and-back

dirvine:
<sup>**[src/bootstrap.rs, line 37 \[r1\]](https://reviewable.io:443/reviews/dirvine/crust/32#-JnSWuGquzZ5tuT7MzvC-r1-37)** ([raw file](https://github.com/dirvine/crust/blob/2cfe8e7d0614e01b0914369687e6f5979f0c16f3/src/bootstrap.rs#L37)):</sup>
http://stackoverflow.com/questions/29570607/is-there-a-good-way-to-convert-a-vect-to-an-array Here ye go :-) 

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/32)

PR #33, opened 2015-04-21
WIP Maid 966 test

mmoadeli:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/33)


src/connection_manager.rs
```diff
@@ -366,4 +411,96 @@ mod test {
...
+                    let ep = end_point.clone();
+                    spawn(move || {
+                        let node = node.lock().unwrap();
+                        let _ = node.conn_mgr.send(ep.clone(), encode(&"message".to_string()));
+                    });
+                }
+            }
+        }
+
+        thread::sleep_ms(100 * NETWORK_SIZE + 10 * MESSAGE_PER_NODE);
```
chandraprakash: Can we avoid sleeps. may be a channel with multiple tx to notify expected count ?

mmoadeli: I am not sure the channel give us any benefits. The thread runs until it reaches a known limit. The sleep can be tuned a bit by adding some logic to investigate the Vec<Stats> to see if all messages are received.

The introduction of channel, requires looping on rx.iter()s for transmissions from thread. Then there is the need for putting time out on how long main thread should wait for the threads to join

dirvine:
Reviewed files:
- src/beacon.rs @ r1

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/33)

PR #34, opened 2015-04-21
WIP: Maid 952 

mmoadeli:
The code delivers what the task asked for. A simple test is also provided to validate it is fine. The test code passes only when the test runs alone. If it runs along with other tests its results depends on whether any other test has taken port 5483 or not

[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/34)


src/connection_manager.rs
```diff
@@ -327,6 +333,21 @@ mod test {
     }
 
 #[test]
+    fn bootstrap() {
+        let (cm1_i, _) = channel();
+        let cm1 = ConnectionManager::new(cm1_i);
+        let cm1_eps = cm1.start_listening(vec![Port::Tcp(0)]).unwrap();
```
chandraprakash: worth calling listen in cm2 as well to test the case when a node fails to acquire 5483 udp port and ignores the failure.

PR #35, opened 2015-04-22
Maid 960 cli crust node

maqi:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/35)

### DIFF
--- a/examples/crust_node.rs
+++ b/examples/crust_node.rs
@@ -223,29 +223,25 @@ fn main() {
   let mut command = String::new();
   loop {
       command.clear();
-      println!("input command >");
+      println!("input command ( stop | connect <Endpoint> | send <Endpoint> <Msg> )>");
+      // stop
+      // connect <Endpoint>
+      // send <Endpoint> <Msg>
       let _ = io::stdin().read_line(&mut command);
-      match command.trim() {
+      let v: Vec<&str> = command.split(' ').collect();
+      println!("main command is : {}", v[0]);
+      match v[0].trim() {
           "stop" => break,
           "send" => {
-              println!("input endpoint to send msg >");
-              let mut endpoint_str = String::new();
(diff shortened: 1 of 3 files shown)


examples/crust_node.rs
```diff
@@ -0,0 +1,214 @@
...
+                      // default_bootstrap = true; }
+        };
+      },
+      None => { println!("No peer address provided, resort to default");
+                default_bootstrap = true; }
+    }
+  }
+
+  // resort to default bootstrapping methods
+  if default_bootstrap && !args.flag_origin {
```
benjaminbollen: Qi, the intention of copying the boolean `flag_bootstrap` into a mutable `default_bootstrap` at L157 was explicitly to be able to mutate it in L174: if a user provides commandline --bootstrap but writes nothing behind it `arg_peer` is `None`, but `flag_bootstrap` is true.  So logically that is "default bootstrap" to me, hence I overwrote it.  Minor detail, but now you could skip all bootstrapping attempts altogether.

maqi: Yes, that's the purpose to allow skip all bootstrapping attempts.
Because for the first node, there is no point trying to bootstrap.

benjaminbollen:
<sup>**[examples/crust_node.rs, line 181 \[r3\]](https://reviewable.io:443/reviews/dirvine/crust/35#-JnV5PutPmLEEZz4IBgV-r3-180)** ([raw file](https://github.com/dirvine/crust/blob/a1ea351bcf4400275ef258c5a53760594be9552c/examples/crust_node.rs#L181)):</sup>
Sorry yes, I misread on the train this morning. it is `flag_origin`.  thanks Qi

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/35)

PR #37, opened 2015-04-22
Fix: bootstrap connection was not being registered

inetic:
And thus was closed shortly after connect.


chandraprakash:
good to go ...

PR #41, opened 2015-04-22
beacon integration : MAID-952

mmoadeli:
It seems that a WIP of this task which uses fixed port has been merged. That may cause trouble if two instances of same test run in parallel. This PR improves on the last PR by adding a function to generate random port numbers.

### DIFF
--- a/src/beacon.rs
+++ b/src/beacon.rs
@@ -183,7 +183,6 @@ pub fn seek_peers(port: Option<Port>) -> Vec<SocketAddr> {
 
 #[cfg(test)]
 mod test {
-    extern crate rand;
     use super::*;
     use std::net::{UdpSocket/*, lookup_addr, lookup_host*/};
     use std::thread;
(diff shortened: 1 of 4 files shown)


src/beacon.rs
```diff
@@ -183,14 +183,22 @@ pub fn seek_peers(port: Option<Port>) -> Vec<SocketAddr> {
 
 #[cfg(test)]
 mod test {
+    extern crate rand;
```
chandraprakash: use extern in lib.rs

src/beacon.rs
```diff
@@ -183,14 +183,22 @@ pub fn seek_peers(port: Option<Port>) -> Vec<SocketAddr> {
...
+    fn next_port() -> u16 {
+            let mut port:u16 = rand::random();
+            port = 1025 + port % 50000;
+            port
+    }
 
 #[test]
     fn test_broadcast() {
-        let port = Port::Tcp(5493);
+        let port = Port::Tcp(next_port());
```
chandraprakash: this is udp port  (beacon_port)

PR #42, opened 2015-04-22
WIP beacon integration : MAID-952

mmoadeli:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/42)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ homepage = "http://www.maidsafe.net"
 [dependencies]
 rustc-serialize = "*"
 cbor = "*"
+rand = "*"
 
 [[example]]
 name = "beacon_client"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ homepage = "http://www.maidsafe.net"
 [dependencies]
 rustc-serialize = "*"
 cbor = "*"
-rand = "*"
 
 [[example]]
 name = "beacon_client"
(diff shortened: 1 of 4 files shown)


mmoadeli:
It seems that a WIP of this task which uses fixed port has been merged. That may cause trouble if two instances of same test run in parallel. This PR improves on the last PR by adding a function to generate random port numbers.

dirvine:
Reviewed files:
- Cargo.toml @ r1

---

<sup>**[src/beacon.rs, line 193 \[r1\]](https://reviewable.io:443/reviews/dirvine/crust/42#-JnYUgauWFWWhIWOWeIv)** ([raw file](https://github.com/dirvine/crust/blob/15f1df60dc128f5abdea1a946bf55b2438eeb44b/src/beacon.rs#L193)):</sup>
might be better to initialise a static var and increment with every call

---

<sup>**[src/connection_manager.rs, line 340 \[r1\]](https://reviewable.io:443/reviews/dirvine/crust/42#-JnYV5Qe0hOfU3YQA14b)** ([raw file](https://github.com/dirvine/crust/blob/15f1df60dc128f5abdea1a946bf55b2438eeb44b/src/connection_manager.rs#L340)):</sup>
think this duplication can be resolved by moving the tests in question to the /tests/lib.rs 

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/42)

PR #45, opened 2015-04-23
beacon integration : Maid-952

mmoadeli:


src/beacon.rs
```diff
@@ -126,7 +129,7 @@ pub fn listen_for_broadcast(our_listening_address: SocketAddr, port: Option<Port
                 Err(error) => println!("Failed receiving a message: {}", error)
             }
         }});
-        Ok(())
+        Ok(Port::Tcp(used_port))
```
chandraprakash: this is udp port

mmoadeli: at the time of writing of this, Port did not have Udp? Is it added?

PR #51, opened 2015-04-24
Test events - NewMessage / NewConnection / LostConnection - Maid 966

mmoadeli:
The test does not pass as the number of messages received is half the one expected. I would be more than happy to investigate it.
also the number of NewConnection events are half the number it should be.

[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/51)


mmoadeli:
It seems that one of my assumption has been wrong. Crust is fine, the test must be changed slightly.

dirvine:
Ok Mahmoud can you confirm when you feel this is complete and working properly

mmoadeli:
I will have it finished today David.

dirvine:
Excellent, just ping me when you have. Thx

mmoadeli:
sure David

mmoadeli:
David, The test changed. It fails/passes intermittently. I can't tell now what is the problem. 
I have a network of N nodes where each node is connected to N-1 node.
What I witnessed is if each node tries to send M messages to N-1 connected nodes, the number of messages failing to deliver is multiple of M. It is likely that one of nodes for some reason is unable to send. I can work on it on weekend.

mmoadeli:
David, It seems that sending message before the connection being established was the problem. adding few ms after connect made it work.

dirvine:
OK Mahmoud, nice one we will look into it, but this PR has conflicts and cannot be merged.

dirvine:
Reviewed files:
- src/connection_manager.rs @ r3

---

<sup>**[src/connection_manager.rs, line 550 \[r3\]](https://reviewable.io:443/reviews/dirvine/crust/51#-Jnr3dQROALBs6OuwHYp)** ([raw file](https://github.com/dirvine/crust/blob/93ec7752f57b4645fcf02880be0b5b4454098ea2/src/connection_manager.rs#L550)):</sup>
It would be much nicer not to need a sleep but as  a temp solution for now it may be OK. If you can think of a better way before tomorrow great, otherwise we will try and get somebody else to take a look as well, which helps everyone. Nice to see it working though. 

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/51)

dirvine:
Reviewed files:
- src/connection_manager.rs @ r4

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/51)

dirvine:
@mmoadeli  can you please fix conflicts

mmoadeli:
David, I fix it later today if it is ok.

mmoadeli:
It is merged

mmoadeli:
Merge conflicts again are fixed

dirvine:
Reviewed files:
- src/connection_manager.rs @ r7

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/51)

PR #54, opened 2015-04-27
Connect manager bench mark test

maqi:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/54)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
 - cd ../..
 script:
 - cargo build --verbose
-- cargo test --verbose
+- RUST_TEST_THREADS=1 cargo test
 after_success: |
   [ $TRAVIS_BRANCH = master ] &&
   [ $TRAVIS_PULL_REQUEST = false ] &&
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "crust"
-version = "0.0.5"
+version = "0.0.51"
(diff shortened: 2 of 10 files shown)


maqi:
This is to replace the PR : https://github.com/dirvine/crust/pull/12

dirvine:
@maqi  Can you please look at conflicts.

inetic:
Hi ya, I'm gonna just sum up what we've discussed over the private rust channel here. For the benchmark, I don't think it does anything useful. Seems, what it does, it creates two CM nodes and then periodically calls the `connect` method from one to the other, I don't think this is something sensible, nor something we would like to benchmark. On top, there is a sleep_ms in the benchmark which I think reduces its value further.

In the future we might want to add benchmark that tests how quickly a set of nodes can interconnect, but due to limitations in rust's current TCP stack (unable to break from blocking calls) this would just crash the bench so I don't think we should attempt it right now (though we might want to have a crashing test case just to remind us of the limitations).

As discussed and agreed upon with @maqi and Viv, a sensible benchmark right now for us would test a throughput between two nodes.

maqi:
The merging conflict has been resolved.
I have also changed the bench test to measure the throughput between two nodes.

dirvine:
Reviewed files:
- benches/connection_manager.rs @ r5
- src/connection_manager.rs @ r5

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/54)

PR #80, opened 2015-05-01
Proof of feasibility Wireshark lua plugin

ned14:
Not actually for merging this, so please close the pull request once review is complete. This is a proof of feasibility for a lua wireshark plugin for crust. A long list of things need fixing in crust before progressing more here makes sense, but it shows it works via heuristic and well at that.

[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/80)

### DIFF
--- a/src/tcp_connections.rs
+++ b/src/tcp_connections.rs
@@ -23,7 +23,7 @@ use std::thread::spawn;
 use std::marker::PhantomData;
 use rustc_serialize::{Decodable, Encodable};
 use std::sync::mpsc;
-use std::sync::mpsc::{Receiver};
+use std::sync::mpsc::Receiver;
 
 pub type InTcpStream<T> = Receiver<T>;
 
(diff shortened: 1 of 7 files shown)


wireshark-plugin/crust.lua
```diff
@@ -0,0 +1,86 @@
...
+    -- so it might be better to just call the function directly instead of doing
+    -- this, but this script is used for testing and this tests the call() function
+    crust.dissector(tvbuf,pktinfo,root)
+
+    -- since this is over a transport protocol, such as UDP, we can set the
+    -- conversation to make it sticky for our dissector, so that all future
+    -- packets to/from the same address:port pair will just call our dissector
+    -- function directly instead of this heuristic function
+    -- this is a new attribute of pinfo in 1.11.3
+    pktinfo.conversation = crust
```
dirvine: nice!

src/tcp_connections.rs
```diff
@@ -30,6 +30,9 @@ pub type InTcpStream<T> = Receiver<T>;
 pub type TcpReader<T> = Receiver<T>;
 pub type TcpWriter<T> = OutTcpStream<T>;
 
+/// Magic used to identify a crust packet
+pub const CRUST_PACKET_HEADER : [u8; 4] = [0x63, 0x72, 0x75, 0x31];  // "cru1"
```
dirvine: So we could use this in our tracer protocol or provide a mechanism for decrypting to get this? I like the tracer idea as it means people can develop in any p2p protocol with any encryption or obfuscation and switch to tracer for test networks. As p2p is be nature distributed getting keys means try decrypt with every key, so a tracer is a neat idea. Crust then allows an automatic multi protocol NAT Traversing solution for p2p networks and handles and try as many protocols that it can in order of efficiency. So we would need a debug type flag to allow crust to go into force tracer mode.

ned14: I think you just need to annotate the initial discovery bootstrap part. I believe the wireshark plugin can tell wireshark that the next step in some sequence is a packet of some protocol on some port, so if it can decode what happens next it simply tells wireshark. Wireshark then should construct a threaded conversation for you.

dirvine: That would be neat indeed.

PR #83, opened 2015-05-05
Guid for beacon

inetic:
To avoid sending beacon broadcast to ourselves.

Should be merged only after Fraser's https://github.com/dirvine/crust/pull/82 is merged

[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/83)

### DIFF
--- a/examples/crust_node.rs
+++ b/examples/crust_node.rs
@@ -38,7 +38,7 @@ use std::io;
 use std::io::Write;
 use std::net::SocketAddr;
 use std::str::FromStr;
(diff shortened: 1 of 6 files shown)


dirvine:
Reviewed files:
- examples/crust_node.rs @ r1
- src/beacon.rs @ r2
- src/lib.rs @ r1
- src/tcp_connections.rs @ r1
- src/transport.rs @ r1

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/83)

PR #87, opened 2015-05-05
Fixed usage of beacon GUID.

Fraser999:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/87)


dirvine:
Reviewed files:
- src/beacon.rs @ r1
- src/tcp_connections.rs @ r2

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/87)

PR #89, opened 2015-05-06
Deny undesirable coding patterns

rossmuir:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/89)

Issue #84, opened 2015-05-05
`crust_node` example frequently fails to stop cleanly on Windows


Fraser999:
Entering `stop` should stop the example exe, but on Windows, frequently it just results in the exe hanging.  If we then attempt to kill it using ctrl+C, the exe appears to have stopped (in that control is returned to the console), but it is actually still running (can be seen and killed via Task Manager).

PR #90, opened 2015-05-06
Remove warnings

Fraser999:
This includes Ross' PR #89

I've got an instance of `#[allow(dead_code)]` at the TCP `close` function.  It's needed to avoid the tests hanging, but it's telling that it's unused in the main codebase.  This could well be the cause of the hanging thread causing the Windows issue #84.

Once I get to the bottom of the issue I expect the `close` function will be getting used then, and the `#[allow...]` can go away.

There are also many places where I'm ignoring the result of `thread::spawn` calls, but again I'll be looking at these while debugging the Windows issue.

[Review on Reviewable](https://reviewable.io/reviews/dirvine/crust/90)

### DIFF
--- a/examples/crust_node.rs
+++ b/examples/crust_node.rs
@@ -29,10 +29,10 @@ extern crate time;
 
 use core::iter::FromIterator;
 use docopt::Docopt;
-use rand::random;
-use rand::Rng;
+// use rand::random;
+// use rand::Rng;
 use rustc_serialize::{Decodable, Decoder};
-use std::cmp;
+// use std::cmp;
 use std::sync::mpsc::channel;
 use std::io;
 use std::io::Write;
@@ -109,13 +109,13 @@ impl Decodable for PeerEndpoint {
     }
 }
 
-fn generate_random_vec_u8(size: usize) -> Vec<u8> {
-    let mut vec: Vec<u8> = Vec::with_capacity(size);
-    for _ in 0..size {
-        vec.push(random::<u8>());
-    }
-    vec
-}
+// fn generate_random_vec_u8(size: usize) -> Vec<u8> {
+//     let mut vec: Vec<u8> = Vec::with_capacity(size);
+//     for _ in 0..size {
+//         vec.push(random::<u8>());
+//     }
+//     vec
+// }
 
 fn print_input_line() {
     print!("Enter command (stop | connect <endpoint> | send <endpoint> <message>)>");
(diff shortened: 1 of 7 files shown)


dirvine:
Reviewed files:
- examples/crust_node.rs @ r1
- src/beacon.rs @ r1
- src/bootstrap.rs @ r1
- src/connection_manager.rs @ r1
- src/lib.rs @ r1
- src/tcp_connections.rs @ r1
- src/transport.rs @ r1

---

<sup>**[src/bootstrap.rs, line 40 \[r1\]](https://reviewable.io:443/reviews/dirvine/crust/90#-JodHug_I_tRgf9yK3MR)** ([raw file](https://github.com/dirvine/crust/blob/51642123129d94d11bd18cf05d508229cc441880/src/bootstrap.rs#L40)):</sup>
Excellent

---

<sup>**[src/connection_manager.rs, line 743 \[r1\]](https://reviewable.io:443/reviews/dirvine/crust/90#-JodI9xCvDeha8VZ3-KN)** ([raw file](https://github.com/dirvine/crust/blob/abe18ec4a4a78be98b39ead3ffaf75b881350600/src/connection_manager.rs#L743)):</sup>
Do you do this to force a join ? seems like a neat trick

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/90)

dirvine:
<sup>**[src/lib.rs, line 22 \[r1\]](https://reviewable.io:443/reviews/dirvine/crust/90#-JodIKG5agxrPx0Fjfe8)** ([raw file](https://github.com/dirvine/crust/blob/51642123129d94d11bd18cf05d508229cc441880/src/lib.rs#L22)):</sup>
Every lib should do this now Ross

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/90)

Fraser999:
<sup>**[src/connection_manager.rs, line 743 \[r1\]](https://reviewable.io:443/reviews/dirvine/crust/90#-JodI9xCvDeha8VZ3-KN)** ([raw file](https://github.com/dirvine/crust/blob/abe18ec4a4a78be98b39ead3ffaf75b881350600/src/connection_manager.rs#L743)):</sup>
No - this is an example of one of the places where we just ignore the return and don't wait for the thread to join.  `thread::scoped` would cause the thread to join when the returned value is destroyed.

These are the red flags I'll be trying to get my head round while debugging #84

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/crust/90)

PR #108, opened 2015-05-15
parallel attempts to bootstrap

bcndanos:
Here you go!

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/108)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ rustc-serialize = "*"
 sodiumoxide = "0.0.4"
 term = "*"
 time = "*"
+asynchronous = "*"
 
 [[example]]
 name = "simple_sender"
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -27,6 +27,8 @@ use bootstrap::{BootStrapHandler, BootStrapContacts, Contact, PublicKey};
 use transport;
 use transport::{Endpoint, Port};
 
+use asynchronous::{Deferred,ControlFlow};
+
 /// Type used to represent serialised data in a message.
 pub type Bytes = Vec<u8>;
 
@@ -295,19 +297,28 @@ impl ConnectionManager {
     }
 
     fn bootstrap_off_list(&self, bootstrap_list: Vec<Endpoint>) -> io::Result<Endpoint> {
+        let mut vec_deferred = vec![];
         for endpoint in bootstrap_list {
-            match transport::connect(endpoint.clone()) {
-                Ok(trans) => {
-                    let ep = trans.remote_endpoint.clone();
-                    let _ = try!(handle_connect(self.state.downgrade(), trans,
-                                                self.beacon_guid_and_port.is_some()));
-                    return Ok(ep)
-                },
-                Err(_) => continue,
-            }
+            let state_cloned = self.state.clone();
+            let beacon_guid_and_port_is_some = self.beacon_guid_and_port.is_some();
+            vec_deferred.push(Deferred::new(move || {
+                match transport::connect(endpoint.clone()) {
+                    Ok(trans) => {
+                        let ep = trans.remote_endpoint.clone();
+                        let _ = try!(handle_connect(state_cloned.downgrade(), trans,
+                                                    beacon_guid_and_port_is_some ));
+                        return Ok(ep)
+                    },
+                    Err(e) => Err(e),
+                }
+            }));
         }
-        // FIXME: The result should probably be Option<Endpoint>
-        Err(io::Error::new(io::ErrorKind::Other, "No bootstrap node got connected"))
+        let res = Deferred::first_to_promise(1,false,vec_deferred, ControlFlow::ParallelLimit(15)).sync();
+        match res {
+            Ok(v) => Ok(v[0].clone()),            
+            Err(_) => Err(io::Error::new(io::ErrorKind::Other, "No bootstrap node got connected"))
+        }
+        // FIXME: The result should probably be Option<Endpoint> 
     }
 
     fn listen(&self, port: &Port) -> io::Result<Endpoint> {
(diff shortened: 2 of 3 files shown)


maidsafe-highfive:
Thanks for the pull request, and welcome! The MaidSafe team is excited to review your changes, and you should hear from @ned14 (or someone else) soon.

If any changes to this PR are deemed necessary, please add them as extra commits. This ensures that the reviewer can see what has changed since they last reviewed the code. The way Github handles out-of-date commits, this should also make it reasonably obvious what issues have or haven't been addressed. Large or tricky changes may require several passes of review and changes.

Please see [CONTRIBUTING.md](https://github.com/maidsafe/maidsafe/blob/master/CONTRIBUTING.md) for more information.

chandraprakash:
Thanks a lot Mike. We will have this reviewed and merged. Great work !!

bcndanos:
There's a problem with this PR. Don't merge.

Executing the crust_node example 

Listening for new connections on Tcp(V4(127.0.1.1:53230)), and listening for UDP broadcast on port 9999.
thread '<main>' panicked at 'index out of bounds: the len is 0 but the index is 0', /home/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-linux/build/src/libcollections/vec.rs:1349
An unknown error occurred

chandraprakash:
Yeah I noticed that too :-)

bcndanos:
I think know it's fixed. Sorry! :|

chandraprakash:
Hi Mike, 
Must say very nice work.  I tried your PR and it exposed a bug in Crust code  #109.  It will be handy if we merge your PR after this bug is fixed.  
Thanks a lot for your help.

dirvine:
r? @ned14 

---

Review status: 0 of 3 files reviewed, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/108)

ned14:
The logic seems fine, but I think surely it would be better if each successfully bootstrapped node was used instead of throwing away all after the first success. I'll look into that here locally.

---

Review status: all files reviewed, all discussions resolved, some commit checks failed.
Reviewed files:
- Cargo.toml @ r1
- src/lib.rs @ r1

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/108)

dirvine:
Ah good point, we do not trust nodes at all so upper layers (routing in our case) should really ask a whole bunch of bootstrap nodes to get info for them, prior to them getting on the network. This way upper layers can filter replies and ensure they are happy.  
The upper layers may need to identify the bootstrap nodes by some id the upper layer uses as the overlay.  This may add complexity as upper layers need to speak individually to each bootstrap node to get this info. So from that perspective if we keep all bootstrap guys and upper layers can negotiate with them all and decide to parallel send info then cool.  Then we can request a node is promoted from bootstrap to connection. Upper layers can then also request certain bootstrap nodes be dropped when they are happy to do so. 
In summary I vote we keep all bootstrap connections, but tag them as such and allow upper payer to do what it wishes with them for as long as they are there. 

---

Review status: all files reviewed, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/108)

bcndanos:
Review status: all files reviewed, 1 unresolved discussion, some commit checks failed.

---

<sup>**[src/connection_manager.rs, line 316 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/108#-Jpr1V2GQvNajal2vR9T)** ([raw file](https://github.com/maidsafe/crust/blob/ee365bfebb278635226384c811012b094dadb429/src/connection_manager.rs#L316)):</sup>
I don't know exactly what you want. But, you can use ' Deferred::vec_to_promise(vec_deferred, ControlFlow::ParallelLimit(15)) ' to execute all the conections instead of the first. In that case, the program will be blocked in this process. Or you can pass as first parameter a number N instead of 1 to stop at the firsts N tasks with Ok. Anyway, the next match has to catch the Err(e), because if there's only one error the return is an Err<Vec<Result<T,E>>>, and you have to iter over the vector to get the OK results.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/108)

ned14:
Review status: all files reviewed, 1 unresolved discussion, some commit checks failed.
Reviewed files:
- Cargo.toml @ r1

---

<sup>**[Cargo.toml, line 20 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/108#-JprCgw51maggocAUCg_)** ([raw file](https://github.com/maidsafe/crust/blob/ee365bfebb278635226384c811012b094dadb429/Cargo.toml#L20)):</sup>
On Windows I'm seeing an error after I merge this pull request:

G:\rust-crust>cargo test
    Updating registry `https://github.com/rust-lang/crates.io-index`
cyclic package dependency: package `num_cpus v0.2.5` depends on itself

So this pull request appears to break crust on Windows?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/108)

bcndanos:
Review status: all files reviewed, 1 unresolved discussion, some commit checks failed.

---

<sup>**[Cargo.toml, line 20 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/108#-JprCgw51maggocAUCg_)** ([raw file](https://github.com/maidsafe/crust/blob/ee365bfebb278635226384c811012b094dadb429/Cargo.toml#L20)):</sup>
Try ' cargo update'. If it doesn't work, try to delete your 'target' directory.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/108)

ned14:
Review status: all files reviewed, 1 unresolved discussion, some commit checks failed.

---

<sup>**[Cargo.toml, line 20 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/108#-JprCgw51maggocAUCg_)** ([raw file](https://github.com/maidsafe/crust/blob/ee365bfebb278635226384c811012b094dadb429/Cargo.toml#L20)):</sup>
Sigh. Yes that worked. Sorry about the noise.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/108)

PR #121, opened 2015-05-29
MAID-1075 and MAID-1122 fix to make listening sockets never have an IP address

ned14:
I'll need to keep adding debug printing until I figure out what's going wrong here. So far it looks like somewhere I'm failing to set a port number, and therefore bootstrapping is done to the wrong location.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/121)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ after_success: |
   ghp-import -n target/doc &&
   git push -fq https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git gh-pages;
   cargo clean;
-  RUST_TEST_TASKS=1 RUST_TEST_THREADS=1 cargo test --no-run;
+  cargo test --no-run;
   kcov --coveralls-id=$TRAVIS_JOB_ID --exclude-pattern=/.cargo target/kcov target/debug/crust-*;
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ after_success: |
   ghp-import -n target/doc &&
   git push -fq https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git gh-pages;
   cargo clean;
-  cargo test --no-run;
+  RUST_TEST_TASKS=1 RUST_TEST_THREADS=1 cargo test --no-run;
   kcov --coveralls-id=$TRAVIS_JOB_ID --exclude-pattern=/.cargo target/kcov target/debug/crust-*;
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,8 @@ after_success: |
   ghp-import -n target/doc &&
   git push -fq https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git gh-pages;
   cargo clean;
-  RUST_TEST_TASKS=1 RUST_TEST_THREADS=1 cargo test --no-run;
-  kcov --coveralls-id=$TRAVIS_JOB_ID --exclude-pattern=/.cargo target/kcov target/debug/crust-*;
+  cargo test --no-run;
+  RUST_TEST_TASKS=1 RUST_TEST_THREADS=1 kcov --coveralls-id=$TRAVIS_JOB_ID --exclude-pattern=/.cargo target/kcov target/debug/crust-*;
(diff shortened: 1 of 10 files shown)


maidsafe-highfive:
r? @chandraprakash

ned14:
Windows fails to build, as expected due to lack of ifaddr(). However from the CI it looks like my code makes assumptions about Rust only valid for rustc 1.2.0 nightly. Should I rebase the code onto Rust 1.1 or 1.0?

dirvine:
Perhaps worth adding `RUST_TEST_TASKS=1` to the line that has   `RUST_TEST_THREADS=1 cargo test` to cover these versions (also test is applicable)

ned14:
The OS X and Windows build runners will need fixing with RUST_TEST_TASKS. Otherwise looks good.

rossmuir:
@ned14 I will update these and rerun

ned14:
Cool. That's now all green except on Windows which is MAID-1122

dirvine:
Review status: 3 of 11 files reviewed, 2 unresolved discussions, some commit checks failed.
Reviewed files:
- README.md @ r3
- src/beacon.rs @ r2
- src/bootstrap.rs @ r2

---

<sup>**[Cargo.toml, line 3 \[r4\]](https://reviewable.io:443/reviews/maidsafe/crust/121#-JqoahxMHH3SQplQYInf)** ([raw file](https://github.com/maidsafe/crust/blob/1f2d7515beb683a0497c0f44ce5f51af39f4e2f5/Cargo.toml#L3)):</sup>
I think we should make this version change after windows is also working and then as a separate version bump PR where we can tag git and also push a new crate. 

---

<sup>**[src/bootstrap.rs, line 281 \[r4\]](https://reviewable.io:443/reviews/maidsafe/crust/121#-JqobMDpu-1qHRzi4X1a)** ([raw file](https://github.com/maidsafe/crust/blob/1f2d7515beb683a0497c0f44ce5f51af39f4e2f5/src/bootstrap.rs#L281)):</sup>
We should perhaps include the name/path of the file in this message to allow users to fix this and restart.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/121)

dirvine:
Overall this is nice work.  Thx Niall, a couple of small issues, if we can address these then this is good for merge and when Windows support is added in the other PR then even better, we will give it a thorough test. I think the GUID is a nice approach 

---

Review status: 6 of 11 files reviewed, 2 unresolved discussions, some commit checks failed.
Reviewed files:
- .travis.yml @ r3
- Cargo.toml @ r2
- src/connection_manager.rs @ r4

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/121)

dirvine:
@ned14 I think when you get MAID-1122 done this is good to merge with that and  we can go for the large PR in this case. As the win part will be a commit anyway it will be fine to review like that.

dirvine:
Very nice @ned14  This feels very solid indeed. Thx for that one, not simple with all the FFI, but worth it.

---

Review status: all files reviewed, 2 unresolved discussions, all commit checks successful.
Reviewed files:
- .travis.yml @ r7
- src/bootstrap.rs @ r6
- src/getifaddrs.rs @ r6
- src/lib.rs @ r5
- src/transport.rs @ r2

---

<sup>**[src/getifaddrs.rs, line 96 \[r7\]](https://reviewable.io:443/reviews/maidsafe/crust/121#-Jqua1mjnPKFtGIwfvIi)** ([raw file](https://github.com/maidsafe/crust/blob/37e23832942e096a0fc62b9b8c9979ad22a9a604/src/getifaddrs.rs#L96)):</sup>
nice, I learned more there, very cool.

---

<sup>**[src/getifaddrs.rs, line 118 \[r7\]](https://reviewable.io:443/reviews/maidsafe/crust/121#-JquaGjJyL3JWf74ebgQ)** ([raw file](https://github.com/maidsafe/crust/blob/37e23832942e096a0fc62b9b8c9979ad22a9a604/src/getifaddrs.rs#L118)):</sup>
We could look at using an iterator here, but for later. This is good for now without doubt.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/121)

PR #123, opened 2015-06-04
Add new start_listening2 API for Routing.

ned14:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/123)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "crust"
-version = "0.0.61"
+version = "0.0.60"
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @chandraprakash

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "crust"
-version = "0.0.60"
+version = "0.0.61"
```
chandraprakash: Can you drop the version bump please. Need to do in a separate pull request.

PR #133, opened 2015-06-12
Code cleanup

vinipsmaker:
Increasing the use of monadic combinators to improve readability.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/133)

### DIFF
--- a/src/beacon.rs
+++ b/src/beacon.rs
@@ -131,28 +131,22 @@ impl BroadcastAcceptor {
         }));
 
         let result = udp_listener_thread.join();
-        match result {
-            Ok(_) => (),
-            Err(e) => {
-                // Connect to the TCP acceptor to allow its thread to join.
-                let _ = TcpStream::connect(("127.0.0.1", self.tcp_listener_port));
-                let _ = tcp_acceptor_thread.join();
-                // Send a ping back to the UDP socket which sent the stop request.
-                match socket_receiver.recv() {
-                    Ok(requester) => {
-                        let sent_size = try!(self.socket.send_to(&[1u8; 1], requester));
-                        debug_assert!(sent_size == 1);
-                    },
-                    Err(_) => (),
-                }
-                return Err(e);
-            },
+        if let Err(e) = result {
+            // Connect to the TCP acceptor to allow its thread to join.
+            let _ = TcpStream::connect(("127.0.0.1", self.tcp_listener_port));
+            let _ = tcp_acceptor_thread.join();
+            // Send a ping back to the UDP socket which sent the stop request.
+            if let Ok(requester) = socket_receiver.recv() {
+                let sent_size = try!(self.socket.send_to(&[1u8; 1], requester));
+                debug_assert!(sent_size == 1);
+            }
+            return Err(e);
         };
         let _ = tcp_acceptor_thread.join();
 
         match transport_receiver.recv() {
             Ok(transport) => Ok(transport),
-            Err(e) => return Err(io::Error::new(io::ErrorKind::BrokenPipe, e.description())),
+            Err(e) => Err(io::Error::new(io::ErrorKind::BrokenPipe, e.description())),
         }
     }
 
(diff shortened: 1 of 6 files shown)


maidsafe-highfive:
Thanks for the pull request, and welcome! The MaidSafe team is excited to review your changes, and you should hear from @ned14 (or someone else) soon.

If any changes to this PR are deemed necessary, please add them as extra commits. This ensures that the reviewer can see what has changed since they last reviewed the code. The way Github handles out-of-date commits, this should also make it reasonably obvious what issues have or haven't been addressed. Large or tricky changes may require several passes of review and changes.

Fraser999:
Hi Vinícius,

Good stuff indeed - thanks very much for this! It's a pretty big payback :-)

All the best,
Fraser.

ned14:
Review status: all files reviewed, 1 unresolved discussion, all commit checks successful.
Reviewed files:
- src/beacon.rs @ r1
- src/bootstrap_handler.rs @ r2
- src/connection_manager.rs @ r1
- src/getifaddrs.rs @ r1
- src/tcp_connections.rs @ r2
- src/transport.rs @ r1

---

<sup>**[src/bootstrap_handler.rs, line 56 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/133#-JrbiyKbjd5LCDet_5-f)** ([raw file](https://github.com/maidsafe/crust/blob/6159fdf9e5457b5ce607ae3be9e9ce225e0931d0/src/bootstrap_handler.rs#L56)):</sup>
How is this returning None if the String::from_utf8 fails?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/133)

vinipsmaker:
> Just compiled without .iter() without problems. Release notes about this change can be found here:
>
> - https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-100-alpha2-february-2015

Thanks for the link, @oblitum. I wasn't aware of all behaviours. I'm aware now, but the code still looks fine (`contacts` is taken by copy, so no hidden side-effects).

vinipsmaker:
Sorry for the possible mail spams you guys got because of me, but these highly spread notifications from GitHub tricked me to make comments on the wrong places about the wrong concerns.

vinipsmaker:
Review status: all files reviewed, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/bootstrap_handler.rs, line 56 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/133#-JrbiyKbjd5LCDet_5-f)** ([raw file](https://github.com/maidsafe/crust/blob/6159fdf9e5457b5ce607ae3be9e9ce225e0931d0/src/bootstrap_handler.rs#L56)):</sup>
The `ok()` member-function converts an `Result<T, E>` to an `Option<T>`. This is exactly what the outer `match` from the previous code was doing.

Then we have the `Ok` case (look the previous code), where it's still possible to return `None`. If we only had the `Some` possibility within the `Ok` case (look the previous code), we could have use `map` and let the `map` function wrap the return from our closure into some `Some`. The `and_then` function looks like `map`, but you return an `Option`.

The inner `match` in the previous code where just converting the return from `json::decode(&contacts_str)` from `Result` to `Option`, which exactly what the member-function `ok` does for us. To be fair, the inner `match` was using control flow constructs to immediately return from `parse_contacts`, but we got the exactly same behaviour w/o these  `return`s, given Rust is an "expression-oriented" language.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/133)

dirvine:
@vinipsmaker  You kidding, it's great to have the effort you have made here. thanks very much, it's not spam :-)

dirvine:
Review status: all files reviewed, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/getifaddrs.rs, line 91 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/133#-Jrc7vzVXcw4lrD6uf4k)** ([raw file](https://github.com/maidsafe/crust/blob/6159fdf9e5457b5ce607ae3be9e9ce225e0931d0/src/getifaddrs.rs#L91)):</sup>
Nice

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/133)

ned14:
Review status: all files reviewed, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/bootstrap_handler.rs, line 56 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/133#-JrbiyKbjd5LCDet_5-f)** ([raw file](https://github.com/maidsafe/crust/blob/6159fdf9e5457b5ce607ae3be9e9ce225e0931d0/src/bootstrap_handler.rs#L56)):</sup>
Yep, you're right. I forgot Result.ok() yields an Option. Sorry.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/133)

PR #157, opened 2015-06-22
fix: now it's safe to call ConnectionManager::stop twice

vinipsmaker:
I've had to clean a lot of debugging code before committing the fix, but it's here now.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/157)


maidsafe-highfive:
r? @chandraprakash

chandraprakash:
Nice one Vinícius !!

PR #170, opened 2015-06-25
Reduced printout by CRUST Fixes #169 

### LINKED ISSUE #169: Reduce printout by CRUST
Can we reduce the printout from CRUST, in particular the failed connections on various interfaces; or have a way to disable it;  Every higher layer piles up more printout (Routing is certainly not innocent here), but CRUST really tends to produce the majority and floods the terminal for more actions on higher levels.

chandraprakash:
Reduced printout by CRUST Fixes #169 
Also ConnectionManager::seek_peers() need not to be public.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/170)


src/transport.rs
```diff
@@ -204,7 +204,7 @@ pub fn connect(remote_ep: Endpoint) -> IoResult<Transport> {
                                          remote_endpoint: remote_ep,
                              }})
                 .map_err(|e| {
-                    let _ = writeln!(&mut io::stderr(), "NOTE: Transport connect {} failure due to {}", ep, e);
+                    // let _ = writeln!(&mut io::stderr(), "NOTE: Transport connect {} failure due to {}", ep, e);
```
dirvine: Can we actually remove this line if no longer required please. It's easy enough to add println! statement when debugging if we wish.

chandraprakash: ok

dirvine:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/170)

PR #176, opened 2015-07-01
Proposed Api changes [DO NOT MERGE]

chandraprakash:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/176)

### DIFF
--- a/installer/crust_peer.config
+++ b/installer/crust_peer.config
@@ -1,6 +1,16 @@
-// If cache is missing or corrupt, its overwritten with empty cache file.
+// Program behaviour:
+// If config file is missing or corrupt, the application will terminate
+
 {
-  "contacts": [
+  "preferred_port": {
+    "variant": "Tcp",
+    "fields": [
+      0
+    ]
+  },
+  "override_default_bootstrap_methods": false,
+  "beacon_port": 9999,
+  "hard_coded_contacts": [
     {
       "endpoint": "45.55.207.180:5483"
     },
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -132,8 +132,9 @@ impl ConnectionManager {
         unimplemented!();
     }
 
-    /// Sends a message to specified addresses (endpoint). Returns Ok(()) if the sending might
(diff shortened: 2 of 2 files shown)


maidsafe-highfive:
r? @ned14

inetic:
OKI, I'll just summarize and rephrase my remark from the meeting. In terms of API I think any softare should first have a good one for developers (functions), then for administrators (config files) and lastly for end users (GUI). The main argument for having it as functions in our particular case is testing, I don't think it is appropriate for a test to first modify a config file for the following reasons:

* It increases the complexity of writing tests, while tests should be as simple as possible
* It will not be possible to run multiple tests from the same directory
* Writing to a file may not be possible or desirable on embedded devices

I think we can very easily have the best of both worlds (config file and functions) without increasing complexity. The API I would love to have would split the current `bootstrap` function into two independent functions. One for reading the config file into memory, say `fn read_config_file() -> Config` and the other one for the actual bootstrapping, say `fn bootstrap(config: Config) -> ...`. Then instead of calling `cm.bootstrap()` one would call `cm.bootstrap(read_config_file())` which I think is a very small price for added modularity with an added bonus that we can then test both functions separately (talking about Crust tests, tests in Routing would simplify greately).

ned14:
@inetic I personally have no objection to config-via-config-file if you can supply at runtime some in-memory config file, and there is an easy to use API for modifying that in-memory config file. In particular it makes fuzz testing much easier for those libraries where unvalidated input can be processed.

inetic:
@ned14, one think is that the current API doesn't support supplying a file descriptor to the `bootstrap` method. But even if it did, IMO it's much easier for us developers to just provide the Config structure with syntax highliting and all the goodies to the `bootstrap` method then creating an in memory string and passing it to in-memory config file.

Correction to the signatures in my previous comment: the config reading function should be `fn read_config_file() -> IoResult<Config>`, then one would use it as:

`cm.bootstrap(read_config_file().or_else(Config::make_default()))`

I think some will argue that this is getting complicated, but to that I would argue that this in one line tells the developer what is happening instead of having to look in multiple lines of documentation or worse looking into implementation details.

PR #187, opened 2015-07-14
MAID-1136_new_bootstrap_event  Fixes #92 Fixes #109

### LINKED ISSUE #92: Bootstrap method should return multiple connected endpoints
Crust's bootstrap method should try to connect to multiple endpoints discovered from multiple sources (responses from multiple beacon listeners and also local bootstrap.cache file).
This means the crust bootstrap method should return vector of *n* endpoints it is connected to.

Details :

If there are X Beacon servers in a LAN and each one of them return Y bootstrap endpoints,
CRUST needs to return after validating each endpoint can be successfully connected a maximum of either X * Y connected endpoints(provided all Y endpoints returned by each beacon server are valid) or introduce a filter to limit connections per beacon server / total returned to the consumer of this library.
(same applies for endpoints from local bootstrap.cache file and any further sources we support in crust library) 

**Needs to be decided: (should 'n' be definable? and if so by which of the following approaches)** Allow consumer of this lib to either specify filter based on cap connections per beacon server / total capped connections returned by CRUST / no filter

### LINKED ISSUE #109: NewConnection events should not include bootstrapped endpoint
Bootstrap method is blocking and returns endpoint it bootstrapped with. Crust also fires a new connection event even for bootstrapped endpoint. We need not to have both options here.

This is happening due to generic handling of all connection including bootstrap connections.
One potential fix will be:  For bootstrap connection, only add connection to the state object if it is selected and returned as bootstrap endpoint. Also don't fire new connection event.

This will become major issue once we start doing parallel bootstrapping (Mike's PR #108). 
As even after bootstrap method returns, crust will continue its parallel bootstrapping and new connection event will be fired along with keeping all tried connection in state object.

chandraprakash:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/187)

### DIFF
--- a/examples/crust_peer.rs
+++ b/examples/crust_peer.rs
@@ -339,7 +339,7 @@ fn main() {
     // Try to bootstrap.  If this fails and we're trying to run the speed test, then fail overall.
     // Otherwise, if no peer endpoints were provided and bootstrapping fails, assume this is
     // OK, i.e. this is the first node of a new network.
-    let connected_peer = match connection_manager.bootstrap(bootstrap_peers.clone(), None) {
+    let connected_peer = match connection_manager.bootstrap_old(bootstrap_peers.clone(), None) {
         Ok(endpoint) => {
             stdout = green_foreground(stdout);
             println!("Bootstrapped to {:?}", endpoint);
--- a/examples/crust_peer.rs
+++ b/examples/crust_peer.rs
@@ -43,6 +43,7 @@ use rand::Rng;
 use rustc_serialize::{Decodable, Decoder};
 use std::cmp;
 use std::sync::mpsc::channel;
+use std::sync::mpsc::Sender;
 use std::io;
 use std::io::Write;
 use std::net::SocketAddr;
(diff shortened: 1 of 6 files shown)


maidsafe-highfive:
r? @ned14

ned14:
Reviewed 5 of 6 files at r1.
Review status: 5 of 6 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/lib.rs, line 22 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/187#-JuG3d2DynhFEdS2LqgX)** ([raw file](https://github.com/maidsafe/crust/blob/860db8947fad9317ce5089b5639c19ca9d084dd2/src/lib.rs#L22)):</sup>
Hmm are these supposed to be commented out?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/187)

chandraprakash:
Also contains work for https://maidsafe.atlassian.net/browse/MAID-1146

chandraprakash:
? @dirvine

PR #188, opened 2015-07-14
Maintains a time ordered list of bootstrap contacts with max size 1500.

s-brian:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/188)

### DIFF
--- a/src/config_utils.rs
+++ b/src/config_utils.rs
@@ -15,6 +15,7 @@
 // Please review the Licences for the specific language governing permissions and limitations
 // relating to use of the SAFE Network Software.
 
+use time;
 use transport::{Endpoint, Port};
 use std::fs::File;
 use std::io::prelude::*;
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @ned14

ned14:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/connection_manager.rs, line 178 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/188#-JuG2mak0H01Guy9XOR7)** ([raw file](https://github.com/maidsafe/crust/blob/01b7b577d046752d75fa6cba825754755d638a13/src/connection_manager.rs#L178)):</sup>
UTP?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/188)

s-brian:
<sup>**[src/connection_manager.rs, line 178 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/188#-JuG2mak0H01Guy9XOR7)** ([raw file](https://github.com/maidsafe/crust/blob/01b7b577d046752d75fa6cba825754755d638a13/src/connection_manager.rs#L178)):</sup>
UTP is not being handled at the moment according the comment under the function signature,  "Currently, only TCP is supported." I guess it should be a new task to add that support and update the code. It's not obvious to me how I would check what endpoint type it should be at the moment. For this task the only change I had to make here was the addition of the timestamp to Contact.  

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/188)

src/config_utils.rs
```diff
@@ -22,10 +23,68 @@ use std::path::PathBuf;
 use std::env;
 use rustc_serialize::json;
 use std::io;
+use rustc_serialize::{Decodable, Decoder, Encodable, Encoder};
+
+#[derive(PartialEq, Eq, Hash, Debug, Clone)]
```
vinipsmaker: This code is failing to compile now, with latest rust nightly:

```
   Compiling crust v0.1.5 (file:///home/vinipsmaker/Projetos/maidsafe/crust)
src/config_utils.rs:30:5: 30:28 error: the trait `core::hash::Hash` is not implemented for the type `time::Tm` [E0277]
src/config_utils.rs:30     pub timestamp: time::Tm,
                           ^~~~~~~~~~~~~~~~~~~~~~~
src/config_utils.rs:28:25: 28:29 note: in expansion of #[derive_Hash]
src/config_utils.rs:28:25: 28:29 note: expansion site
src/config_utils.rs:30:5: 30:28 help: run `rustc --explain E0277` to see a detailed explanation
error: aborting due to previous error
Could not compile `crust`.
```

I can update it, but I'd like to confirm it was working before, so I may investigate how to mimic previous behaviour.

s-brian: Looks like you've not pulled the latest if the file your using has this code in it.

PR #191, opened 2015-07-15
WIP Maid 1142

ned14:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/191)


maidsafe-highfive:
r? @chandraprakash

ned14:
UTP support still needs a lot of debugging in the Rust side of things, but I am out of hours, so here it is. It should merely be a case of figuring out the Rust iterator problem I mentioned on Slack earlier today, other than that it appears to be working very well.

The UTP implementation (written in C++ 14) is very solid and has its own unit tests. The only problem is the Rust stuff.

Note that builds may fail on the CI because the new build command now is:

CXX=g++-4.9 cargo test

... where your C++ compiler does not support C++14

PR #197, opened 2015-07-21
WIP fix 195

vinipsmaker:
**DO NOT MERGE IT**

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/197)

### DIFF
--- a/src/tcp_connections.rs
+++ b/src/tcp_connections.rs
@@ -16,7 +16,7 @@
 // relating to use of the SAFE Network Software.
 
 use std::net::{TcpListener, TcpStream, SocketAddr, Shutdown};
-use std::io::{BufReader, ErrorKind};
+use std::io::{BufReader, Error, ErrorKind};
 use std::io::Result as IoResult;
 use cbor::{Encoder, CborError, Decoder};
 use std::thread;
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @chandraprakash

src/tcp_connections.rs
```diff
@@ -53,7 +53,13 @@ impl <T> Drop for OutTcpStream<T> {
 /// Connect to a peer and open a send-receive pair.  See `upgrade` for more details.
 pub fn connect_tcp<'a, 'b, I, O>(addr: SocketAddr) -> IoResult<(Receiver<I>, OutTcpStream<O>)>
         where I: Send + Decodable + 'static, O: Encodable {
-    Ok(try!(upgrade_tcp(try!(TcpStream::connect(&addr)))))
+    let stream = try!(TcpStream::connect(&addr));
+    if try!(stream.peer_addr()).port() == try!(stream.local_addr()).port() {
```
chandraprakash: With this we restrict same src/dst port connection. As crust doesn't provide src port in it's API, this should be okay.

We need to document the behaviour in API so that upper layer retries if it gets ConnectionRefused error.
Other option is to retry connect internally and it will likely pick a different port.

Will be good to hear from @dirvine

PR #199, opened 2015-07-21
MAID-1264 Config path documentation and updates

chandraprakash:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/199)

### DIFF
--- a/docs/configuration_path.md
+++ b/docs/configuration_path.md
@@ -22,7 +22,7 @@ Cons:
 - Running parallel tests will interfere each other.
 - Multiple instances of same application cannot have customised config option per instance.
 - Any user can override existing `SystemAppSupportDir ` config file by creating `UserAppDir`
- -
+ 
 ####Writing order
 In case the configuration file is missing at all above paths, crust will attempt to create file at following path in same order.
 
--- a/docs/configuration_path.md
+++ b/docs/configuration_path.md
@@ -1,7 +1,7 @@
 ### Configuration file path
 #### Naming convention
-File name needs to be of the format : `[current_executable_name].config`
-For crust_peer example, the config file name will be `crust_peer.config`
+File name needs to be of the format : `[current_executable_name].crust.config`
+For crust_peer example, the config file name will be `crust_peer.crust.config`
 
 
 ####Reading order
@@ -17,12 +17,13 @@ Pro`s:
 - Installers can create the config file at appropriate path with desired name and permissions.
 - Tests can override other paths by placing config file in current exe directory.
 - File name format helps in separating config file of different application.
+- Using `std::env::current_exe` against working directory helps in cleanup as config files gets created on fixed location(s). Example: running test file from multiple location will create multiple config files and cleanup is difficult.
 
 Cons:
-- Running parallel tests will interfere each other.
+- Running parallel tests will interfere each other. [This can be addressed by using memory mapped file.]
 - Multiple instances of same application cannot have customised config option per instance.
-- Any user can override existing `SystemAppSupportDir ` config file by creating `UserAppDir`
- 
+- Any user can override existing `SystemAppSupportDir ` config file by creating `UserAppDir`.[User can only pollute Vaults running under current user only]
+
 ####Writing order
 In case the configuration file is missing at all above paths, crust will attempt to create file at following path in same order.
 
@@ -35,20 +36,24 @@ In case the configuration file is missing at all above paths, crust will attempt
 - If config options are default for an application, its installer can skip installing it and application will auto create appropriate file at best path if it can.
 
 #####Cons:
-- If its a client user app having permissions to write to system dir, it will create system dir and not home dir
- 
+- If its a client user app having permissions to write to system dir, it will create system dir and not home dir. [This is a sysadmin error not a program error.]
+
 
 #### Platform specific paths
 **UserAppDir**
- - *Windows* - `env::var("APPDATA") / CompanyName() / ApplicationName()`
- - *APPLE* - `env::home_dir() / "Library/Application Support" / CompanyName() / ApplicationName()`
- - *Linux* -  `env::home_dir() / ".config" / CompanyName() / ApplicationName()`
+ - *Windows* - `env::var("APPDATA") / ExeName()`
+ - *APPLE* - `env::home_dir() / "Library/Application Support" / ExeName()`
+ - *Linux* -  `env::home_dir() / ".config" / ExeName()`
 
 **SystemAppSupportDir**
- - *Windows* - `env::var("ALLUSERSPROFILE") / CompanyName() / ApplicationName()`
- - *Apple* - `"/Library/Application Support/" / CompanyName() / ApplicationName()`
+ - *Windows* - `env::var("ALLUSERSPROFILE") / ExeName()`
+ - *Apple* - `"/Library/Application Support/" / ExeName()`
  - *Linux* - `"/opt/" / CompanyName() / "sbin"`
 
 ### Bootstrap cache file path
 
+Same conventions as config file will be used for bootstrap cache file as well.
 
+#### Naming convention
+File name needs to be of the format : `[current_executable_name].crust.bootstrap.cache`
+For crust_peer example, the config file name will be `crust_peer.crust.bootstrap.cache`
--- a/docs/configuration_path.md
+++ b/docs/configuration_path.md
@@ -6,9 +6,9 @@ For crust_peer example, the config file name will be `crust_peer.config`
 
 ####Reading order
 On construction of `crust::connection_manager` instance, crust will try to read configuration file from the following path in same order.
-1. Current executable directory: `std::env::current_exe`
-2. Current user's application directory:  `UserAppDir`
-3. Application support directory for all users: `SystemAppSupportDir `
+1. Current executable directory: `std::env::current_exe` 
+2. Current user's application directory:  `UserAppDir` 
+3. Application support directory for all users: `SystemAppSupportDir ` 
 
 #####Pro`s:
 
--- a/docs/configuration_path.md
+++ b/docs/configuration_path.md
@@ -1,0 +1,40 @@
+### Configuration file path
+#### Naming convention
+File name needs to be of the format : `[current_executable_name].config`
+For crust_peer example, the config file name will be `crust_peer.config`
+
+
+####Reading order
+On construction of `crust::connection_manager` instance, crust will try to read configuration file from the following path in same order.
+1. Current executable directory: `std::env::current_exe`
+2. Current user's application directory:  `UserAppDir`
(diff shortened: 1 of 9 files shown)


maidsafe-highfive:
r? @ned14

docs/configuration_path.md
```diff
@@ -0,0 +1,54 @@
+### Configuration file path
+#### Naming convention
+File name needs to be of the format : `[current_executable_name].config`
```
dirvine: app_name.crust.config() ?

docs/configuration_path.md
```diff
@@ -0,0 +1,54 @@
...
+ 
+
+#### Platform specific paths
+**UserAppDir**
+ - *Windows* - `env::var("APPDATA") / CompanyName() / ApplicationName()`
+ - *APPLE* - `env::home_dir() / "Library/Application Support" / CompanyName() / ApplicationName()`
+ - *Linux* -  `env::home_dir() / ".config" / CompanyName() / ApplicationName()`
+
+**SystemAppSupportDir**
+ - *Windows* - `env::var("ALLUSERSPROFILE") / CompanyName() / ApplicationName()`
```
dirvine: remove company_name()

docs/configuration_path.md
```diff
@@ -0,0 +1,54 @@
+### Configuration file path
+#### Naming convention
+File name needs to be of the format : `[current_executable_name].config`
+For crust_peer example, the config file name will be `crust_peer.config`
+
+
+####Reading order
+On construction of `crust::connection_manager` instance, crust will try to read configuration file from the following path in same order.
+
+1. Current executable directory: using `std::env::current_exe`
```
dirvine: Must be clearly documented in API docs + example for testing

docs/configuration_path.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+Pro`s:
+- Intermediate layer libraries need not to know anything about networking configuration.
+- Installers can create the config file at appropriate path with desired name and permissions.
+- Tests can override other paths by placing config file in current exe directory.
+- File name format helps in separating config file of different application.
+
+Cons:
+- Running parallel tests will interfere each other.
```
dirvine: Could be lock-able mmap'ed file

docs/configuration_path.md
```diff
@@ -0,0 +1,54 @@
...
+1. Application support directory for all users: `SystemAppSupportDir `
+2. Current user's application directory:  `UserAppDir`
+
+#####Pro`s
+- Application's advance users can modify config easily by altering the file
+- Apps not having permissions for creating `SystemAppSupportDir ` will create `UserAppDir` after failing to create `SystemAppSupportDir `. This means crust stays agnostic of type of application (Client or Vault)
+- If config options are default for an application, its installer can skip installing it and application will auto create appropriate file at best path if it can.
+
+#####Cons:
+- If its a client user app having permissions to write to system dir, it will create system dir and not home dir
```
dirvine: This is a sysadmin error not a prog error.

docs/configuration_path.md
```diff
@@ -0,0 +1,54 @@
...
+ 
+
+#### Platform specific paths
+**UserAppDir**
+ - *Windows* - `env::var("APPDATA") / CompanyName() / ApplicationName()`
+ - *APPLE* - `env::home_dir() / "Library/Application Support" / CompanyName() / ApplicationName()`
```
dirvine: ApplicationName -> ExeName

docs/configuration_path.md
chandraprakash: Noted and will do it in implementation part.

docs/configuration_path.md
```diff
@@ -0,0 +1,59 @@
...
+#### Platform specific paths
+**UserAppDir**
+ - *Windows* - `env::var("APPDATA") / ExeName()`
+ - *APPLE* - `env::home_dir() / "Library/Application Support" / ExeName()`
+ - *Linux* -  `env::home_dir() / ".config" / ExeName()`
+
+**SystemAppSupportDir**
+ - *Windows* - `env::var("ALLUSERSPROFILE") / ExeName()`
+ - *Apple* - `"/Library/Application Support/" / ExeName()`
+ - *Linux* - `"/opt/" / CompanyName() / "sbin"`
```
vinipsmaker: `CompanyName` is still here. What do?

vinipsmaker: [Old C++ code](https://github.com/maidsafe-archive/MaidSafe-Common/blob/f7f935f15a14d64cecd026c116b59fdd57b9079a/include/maidsafe/common/application_support_directories.h#L39) relied on compilation time information. [It doesn't seem this approach would work in Rust](http://doc.crates.io/manifest.html). The closest thing is features, but only allow `present`/`absent` value and it doesn't cross crates boundaries (if crust is a nested/indirect dependency, you just can't pass the required information).

[Qt, which also uses company name, relies on global variables](http://doc.qt.io/qt-5/qcoreapplication.html#organizationName-prop). Global variables aren't so nice in Rust. And this approach isn't nice anyway. I only mentioned Qt, because Qt has its fame when it comes to cross-platform support.

I think [we could just use an envvar](https://doc.rust-lang.org/stable/std/env/fn.set_var.html) to mimic Qt approach nicely in Rust. But I believe @dirvine [wanted to get rid of the `CompanyName` thing entirely](https://github.com/maidsafe/crust/pull/199#discussion_r35105705).

I'll use `"/opt/" / ExeName()` for now and I'll update the code when a decision is reached.

dirvine: Ah should be /var/cache/ExeName() for linux I believe

vinipsmaker: /var/cache isn't not a good idea for configuration files. Files under /var/cache are good candidates for removal if file system is getting full.

systemd docs are a great source to choose an appropriate path. I believe /var/lib would work better.

vinipsmaker: Forgot to add "thought?" on my last comment.

vinipsmaker: After some research, I believe /var/lib is the best option:

["Persistent system data. System components may place private data in this directory"](http://www.freedesktop.org/software/systemd/man/file-hierarchy.html).

I'll go with that and I can revert in case you guys want a different option.

vinipsmaker:
r? vinipsmaker

examples/crust_peer.rs
```diff
@@ -34,7 +34,7 @@ extern crate docopt;
 extern crate rand;
 extern crate term;
 extern crate time;
-extern crate tempdir;
+// extern crate tempdir;
```
vinipsmaker: Should get rid of commented code. VCS can already track unused code.

src/config_utils.rs
```diff
@@ -33,23 +34,98 @@ pub type Contacts = Vec<Contact>;
...
 #[derive(PartialEq, Debug, RustcDecodable, RustcEncodable, Clone)]
 pub struct Config {
-    pub preferred_ports: Vec<Port>,
     pub override_default_bootstrap: bool,
     pub hard_coded_contacts: Contacts,
     pub beacon_port: u16,
 }
 
-pub fn default_config_path() -> io::Result<(PathBuf)> {
+// pub fn default_config_path() -> io::Result<(PathBuf)> {
```
vinipsmaker: should get rid of commented code

src/connection_manager.rs
```diff
@@ -211,34 +205,11 @@ impl ConnectionManager {
...
-                            IpAddr::V4(a) => Endpoint::utp((a, p)),
-                            IpAddr::V6(a) => Endpoint::utp((a, p)),
-                        });
-                    }
-                },
-            }
-        }
-        Ok(endpoints)
+    /// return a vector of listening endpoints
+    pub fn get_listening_endpoints(&self) -> io::Result<(Vec<Endpoint>)> {
```
chandraprakash: Rename get_accepting_endpoints

src/connection_manager.rs
```diff
@@ -780,34 +764,41 @@ mod test {
...
+        pub path: PathBuf
+    }
+
+    impl Drop for TestConfigFile {
+        fn drop(&mut self) {
+            let _ = remove_file(&self.path);
+        }
+    }
+
+    fn make_temp_config(beacon_port: Option<u16>) -> TestConfigFile {
```
chandraprakash: TestConfigFile and TempDir are both used as raii we can get rid of TempDir.
Address this as a part of cleanup

src/lib.rs
```diff
@@ -19,14 +19,14 @@
...
-#![warn(trivial_casts, trivial_numeric_casts, unused, unused_extern_crates, unused_import_braces,
-        unused_qualifications, unused_results, variant_size_differences)]
+// #![forbid(missing_docs, warnings)]
+// #![deny(bad_style, deprecated, drop_with_repr_extern, improper_ctypes, non_shorthand_field_patterns,
+//         overflowing_literals, plugin_as_library, private_no_mangle_fns, private_no_mangle_statics,
+//         raw_pointer_derive, stable_features, unconditional_recursion, unknown_lints,
+//         unsafe_code, unsigned_negation, unused_allocation, unused_attributes,
+//         unused_comparisons, unused_features, unused_parens, while_true)]
+// #![warn(trivial_casts, trivial_numeric_casts, unused, unused_extern_crates, unused_import_braces,
+//         unused_qualifications, unused_results, variant_size_differences)]
```
chandraprakash: uncomment

src/utils.rs
```diff
@@ -0,0 +1,70 @@
...
+        Error::IoError(e)
+    }
+}
+
+fn not_found_error() -> IoError {
+    use std::io::ErrorKind as IoErrorKind;
+    IoError::new(IoErrorKind::NotFound, "No file name component")
+}
+
+#[allow(dead_code)]
```
chandraprakash: remove this

vinipsmaker: but this function might not be used in other platforms we choose to support in the future. the rust compiler would refuse to compile the code for no good reason. are you sure?

dirvine: I would say we can add a platform specific condition in that case `[fg(if not platofrm) "linux"]` type condition

vinipsmaker: [Solved](https://github.com/vinipsmaker/crust/commit/8f57e5998ecb44a7b74ad0e79184676156834a02).

dirvine: Nice one night owl :-)

vinipsmaker: =)

ned14:
I'm still getting email for all pull requests crust related - can you remove me from the crust project please? Thanks. Niall

src/connection_manager.rs
vinipsmaker: @chandraprakash: [Done](https://github.com/maidsafe/crust/pull/214).

PR #210, opened 2015-07-28
fixes #207 get_own_endpoints returns correct port

### LINKED ISSUE #207: ConnectionManager::get_own_endpoints returns 0 as port
CRUSTv0.2.0

when populating the accepting_on endpoints in routing, I find that I get eg 

local_endpoints: [Tcp(V4(127.0.0.1:0)), Tcp(V4(192.168.0.32:0)), Tcp(V4(172.17.42.1:0))]

so on the receiving end of the connect_request message, the connect_manager.connect() call is not effective, and routing can't establish a new connection between the two nodes.

chandraprakash:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/210)

### DIFF
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -473,8 +473,8 @@ impl ConnectionManager {
 
     fn listen(&mut self, port: &Port) {
         let acceptor = transport::new_acceptor(port).unwrap();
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @dirvine

PR #212, opened 2015-07-29
Version change to 0.2.1

vinipsmaker:
Fixes bug, so it's a maintenance release and only PATCH version number
bumps.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/212)


maidsafe-highfive:
r? @chandraprakash

PR #216, opened 2015-07-30
cleanup: transport::{Endpoint, Port} derive from Copy

vinipsmaker:
Rust will clone Port and Endpoint automatically for us.

Acceptor doesn't derive from `Copy` and it makes sense, but I see no reason why
`Endpoint` and `Port` don't derive from `Copy`.

For reference: http://doc.rust-lang.org/std/marker/trait.Copy.html

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/216)


maidsafe-highfive:
r? @dirvine

PR #218, opened 2015-07-30
Fixes #215 added explicit filter_loopback() method to remove loopback address.

### LINKED ISSUE #215: connection_manager::get_own_endpoints() should not return loopback address
`connection_manager::get_own_endpoints()` should not return loopback address.
There is no use case found yet to have loopback address returned by crust to upper layers.
Currently it is causing noise in upper layer and we decided to not return loopback address on calls to 
`connection_manager::get_own_endpoints()`

chandraprakash:
Added explicit filter_loopback() method to remove loopback address.
Tried to keep filter loopback method separate from getifaddrs to maintain its standard behaviour.
Note: need to cleanup connection manager to use only one call to getifaddrs().
But  getifaddrs() still needs to be called each time user calls ` get_own_endpoints()`.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/218)

### DIFF
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -26,7 +26,7 @@ use std::net::{IpAddr, SocketAddr, SocketAddrV4};
 use beacon;
 use bootstrap_handler::{BootstrapHandler, parse_contacts};
 use config_utils::{Config, Contact, Contacts, read_config_file};
-use getifaddrs::getifaddrs;
+use getifaddrs::{getifaddrs, filter_loopback};
 use transport;
 use transport::{Endpoint, Port};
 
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @dirvine

dirvine:
Is the failing OSX test down to getifaddr not working there ?

chandraprakash:
No its tcp_connections::test::test_multiple_nodes_small_stream failing on a low spec mac.
Need to downscale the test I think.

dirvine:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[src/getifaddrs.rs, line 393 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/218#-JvTfO8FMWczpNG6IDpv)** ([raw file](https://github.com/maidsafe/crust/blob/2d8106a914deb8b1493f86a5c7ccb52f314dbdfc/src/getifaddrs.rs#L393)):</sup>
nice

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/218)

chandraprakash:
@rossmuir  Looks like the test is interfering with other tests .
I can run entire test suite on a mac mini without any issue.

dirvine:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/218)

Issue #223, opened 2015-07-31
Error: use of unstable library feature 'udp_extras': available through the `net2` crate on crates.io


rossmuir:
Unable to compile crust - `rustc 1.3.0-nightly (4d52d7c85 2015-07-30)`

```
Compiling crust v0.2.1 (file:///home/user/MaidSafe-Testing/maidsafe_rust/crust)
src/beacon.rs:198:17: 198:36 error: use of unstable library feature 'udp_extras': available through the `net2` crate on crates.io
src/beacon.rs:198     try!(socket.set_broadcast(true));
                                  ^~~~~~~~~~~~~~~~~~~
<std macros>:1:1: 6:48 note: in expansion of try!
src/beacon.rs:198:5: 198:38 note: expansion site
src/beacon.rs:198:17: 198:36 help: add #![feature(udp_extras)] to the crate attributes to enable
src/lib.rs:33:25: 33:28 error: unused or unknown feature
src/lib.rs:33 #![feature(ip_addr, ip, udp, arc_weak, socket_timeout, duration, negate_unsigned)]
                                      ^~~
src/lib.rs:27:29: 27:44 note: lint level defined here
src/lib.rs:27         unused_comparisons, unused_features, unused_parens, while_true)]
                                          ^~~~~~~~~~~~~~~
src/beacon.rs:198:17: 198:36 error: use of deprecated item: available through the `net2` crate on crates.io
src/beacon.rs:198     try!(socket.set_broadcast(true));
                                  ^~~~~~~~~~~~~~~~~~~
<std macros>:1:1: 6:48 note: in expansion of try!
src/beacon.rs:198:5: 198:38 note: expansion site
src/lib.rs:23:20: 23:30 note: lint level defined here
src/lib.rs:23 #![deny(bad_style, deprecated, drop_with_repr_extern, improper_ctypes, non_shorthand_field_patterns,
                                 ^~~~~~~~~~
error: aborting due to 3 previous errors
Could not compile `crust`.

To learn more, run the command again with --verbose.
   Compiling crust v0.2.1 (file:///home/user/MaidSafe-Testing/maidsafe_rust/crust)
src/beacon.rs:198:17: 198:36 error: use of unstable library feature 'udp_extras': available through the `net2` crate on crates.io
src/beacon.rs:198     try!(socket.set_broadcast(true));
                                  ^~~~~~~~~~~~~~~~~~~
<std macros>:1:1: 6:48 note: in expansion of try!
[...]

PR #224, opened 2015-07-31
fixes using unstable set_broadcast method

mmoadeli:
fixes issue #223 

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/224)

### DIFF
--- a/src/beacon.rs
+++ b/src/beacon.rs
@@ -192,6 +192,7 @@ impl BroadcastAcceptor {
     }
 }
 
+#[allow(deprecated)]
 pub fn seek_peers(port: u16, guid_to_avoid: Option<GUID>) -> Result<Vec<SocketAddr>> {
     // Bind to a UDP socket
     let socket = try!(UdpSocket::bind("0.0.0.0:0"));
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @dirvine

rossmuir:
Is this really the only fix? As this does not move us closer to transitioning to beta / stable?

mmoadeli:
@rossmuir : this is the fix, as long as we want to use unstable features. Using unstable features is not allowed in stable rust.

rossmuir:
@mmoadeli I get this :smile:  but can we not find stable alternatives? > https://github.com/rust-lang/rust/commit/798ce502845435d94856a7b5fed5f703b06506d1 to a novice like me using `#[allow(deprecated)]` feels like just kicking the can down the road?

mmoadeli:
@rossmuir I did not find the stable version of the function. My attempt was to let it compile. I will look for the proper fix, if I am asked so.

Viv-Rajkumar:
@mmoadeli think @rossmuir has a valid point. Lets definitely look for stable alternatives as indicated in the linked commit and update where we can please. Gonna merge this to not have other libraries fail builds for now. can we have this raised as an issue please @rossmuir and assigned to @mmoadeli

Issue #120, opened 2015-05-27
Random Endpoint generation for tests


benjaminbollen:
There is no 'tool' for generating a random endpoint for tests.  Possiblty worth considering.

zacstewart:
Could you clarify this a little? I'm still trying to get a grip on the terminology: what is an endpoint in this context?

In the Ruby world, there are tools like Faker and Forgery which can generate random pieces of data like phone numbers and IP addresses. Are you looking for something like that here?

chandraprakash:
Enpoint is an Enum representing endpoint of supported protocols in crust. 
https://github.com/maidsafe/crust/blob/master/src/transport.rs#L32

You can see an example here: https://github.com/maidsafe/routing/blob/master/src/test_utils/types_util.rs

dirvine:
If you check the utp repo you will find how they use this very similar feature :-) https://github.com/meqif/rust-utp/blob/master/tests/stream.rs

chandraprakash:
There is an implementation already in routing and you need to update it crust.(using Random trait) 
https://github.com/maidsafe/routing/blob/master/src/test_utils/messages_util.rs#L28

inetic:
@benjaminbollen @dirvine, so Routing currently has its own function to generate random Endpoints and Connections, could you confirm that this is OK and we can close this issue or whether you want Crust to export such functionality.

inetic:
I may have accidentally opened this issue myself :-P, closing again.

PR #252, opened 2015-08-17
handling config & bootstrap

mmoadeli:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/252)


maidsafe-highfive:
r? @s-brian

mmoadeli:
The work on bootstrap handler is done and tests are passing. A little clean up is required. Also, the test for config file also need to be added.

mmoadeli:
? @Fraser999

mmoadeli:
bootstrap handler is improved slightly. Despite my initial investigations, the config file did not conform to the flowchart and I spent most of the afternoon fixing and testing config file. 
After talking to David regarding test, I tried to test bootstrap handler outside using test. I found a small bug.
So to sum up config file improved and tested. Bootstrap handler has an issue and must be tested again.

mmoadeli:
bootstrap handler and config files both have been changed slightly.

src/bootstrap_handler.rs
```diff
@@ -102,8 +92,10 @@ impl BootstrapHandler {
     }
 
     pub fn read_bootstrap_file(&mut self) -> io::Result<(Contacts)> {
-        self.file_path = BootstrapHandler::get_file_path();
-        let mut file = try!(File::open(&self.file_path));
+        let mut file = match File::open(&self.file_path) {
```
Fraser999: Not caused by this PR, but this function doesn't `self` to be `mut`.  The same is true for `oldest_contacts` and `get_serialised_contacts`.

src/bootstrap_handler.rs
```diff
@@ -129,8 +121,27 @@ impl BootstrapHandler {
 
     fn write_bootstrap_file(&mut self, mut contacts: Contacts) -> io::Result<()> {
         contacts = contacts.clone().into_iter().unique().collect();
-        self.file_path = BootstrapHandler::get_file_path();
-        let mut file = try!(File::create(&self.file_path));
+        let _ =  fs::create_dir_all(self.file_path.parent().unwrap());
```
Fraser999: Use of `unwrap`.

src/bootstrap_handler.rs
```diff
@@ -129,8 +121,27 @@ impl BootstrapHandler {
...
+                let user_dir = path::PathBuf::from(utils::user_app_dir().unwrap()
+                                    .join(self.file_path.file_name().unwrap()));
+                if self.file_path != user_dir {
+                    match fs::create_dir_all(user_dir.parent().unwrap()) {
+                        Ok(_) => {
+                            let file =  try!(File::create(&user_dir));
+                            self.file_path = user_dir;
+                            file
+                        },
+                        Err(e) => return Err(e)
```
Fraser999: Since we're just returning the same error as is generated by `create_dir_all`, shouldn't we just be doing a `try!(create_dir_all(...))`?

src/bootstrap_handler.rs
```diff
@@ -129,8 +121,27 @@ impl BootstrapHandler {
...
+                    }
+                }
```
Fraser999: `else` should be on this line.

src/bootstrap_handler.rs
```diff
@@ -184,6 +194,33 @@ mod test {
         path.parent().unwrap().join(name)
     }
 
+    fn create_contacts() -> Contacts {
+        let addr = net::SocketAddrV4::new(net::Ipv4Addr::new(1,2,3,4), 8080);
```
Fraser999: I thought we already had helper functions to create random addresses?  If so, I think we should use random addresses.  Ideally we should use random types too, i.e. not limited to TCP and IPv4, but those helpers may not be implemented yet.

src/bootstrap_handler.rs
```diff
@@ -184,6 +194,33 @@ mod test {
...
+        let contact  = Contact { endpoint: Endpoint::Tcp(SocketAddr::V4(addr)) };
+        let mut contacts = Contacts::new();
+        contacts.push(contact.clone());
+        contacts
+    }
+
+    #[test]
+    fn bootstrap_read() {
+        match BootstrapHandler::new().as_mut() {
```
Fraser999: If we don't use `&mut self` in `read_bootstrap_file`, we don't need the `as_mut()` here.

src/bootstrap_handler.rs
mmoadeli: Yes. That's better

src/bootstrap_handler.rs
mmoadeli: Ok

src/bootstrap_handler.rs
mmoadeli: that's correct. I used it in the test which I deleted. forgot to make it un-mut

src/bootstrap_handler.rs
mmoadeli: It is borrowed from other test in the same file.

src/bootstrap_handler.rs
```diff
@@ -184,6 +194,33 @@ mod test {
...
+    fn create_contacts() -> Contacts {
+        let addr = net::SocketAddrV4::new(net::Ipv4Addr::new(1,2,3,4), 8080);
+        let contact  = Contact { endpoint: Endpoint::Tcp(SocketAddr::V4(addr)) };
+        let mut contacts = Contacts::new();
+        contacts.push(contact.clone());
+        contacts
+    }
+
+    #[test]
+    fn bootstrap_read() {
```
Fraser999: This whole test could be reduced to a single line: `assert!(BootstrapHandler::new().is_ok());`

src/bootstrap_handler.rs
```diff
@@ -184,6 +194,33 @@ mod test {
...
+        match BootstrapHandler::new().as_mut() {
+            Ok(handler) => println!("{:?}", handler.read_bootstrap_file()),
+            Err(_) => assert!(false)
+        }
+    }
+
+    #[test]
+    fn bootstrap_write() {
```
Fraser999: Again this could be reduced since we can use `unwrap` in tests:

    let mut handler = BootstrapHandler::new().unwrap();
    assert!(handler.write_bootstrap_file(create_contacts()).is_ok());

src/bootstrap_handler.rs
```diff
@@ -219,10 +256,14 @@ mod test {
         let file_path = get_file_path();
         let path = Path::new(&file_path);
 
-        let mut bootstrap_handler = BootstrapHandler::new();
         let file = fs::File::create(&path);
         assert!(file.is_ok());
 
+        let mut bootstrap_handler = match BootstrapHandler::new() {
```
Fraser999: Just `unwrap` here rather than `match`.  Several other similar cases below.

src/bootstrap_handler.rs
mmoadeli: where to read?

src/bootstrap_handler.rs
Fraser999: Although, I think at least in this test (maybe more) we should close the file and re-read, checking that the contents persist.

src/bootstrap_handler.rs
mmoadeli: also, uness read_bootstrap_file becomes immutable, this should be as_mut. I changed read_bootstrap_file to take self as immutable

src/bootstrap_handler.rs
mmoadeli: ok

src/bootstrap_handler.rs
mmoadeli: They are changed

src/bootstrap_handler.rs
mmoadeli: Not in this particular PR. In this file.

Fraser999:
Overall I don't think this is correct.  For example, going by the flowchart, when we create the file, we should be trying first in the sys dir, not the user dir.

We use the same function to get the file path regardless of whether the file exists or not.  I think we need different functions to get the file path depending on whether we're trying to read, write (i.e. update), or create the file.  Or we use the same function, but it needs to know the context in which it's being called (i.e. it'd need to know whether we're currently trying to read, write or create the file).

mmoadeli:
thanks @Fraser999 . I address the issue I need to change the creation path

src/bootstrap_handler.rs
Fraser999: Both actually.

src/bootstrap_handler.rs
Fraser999: So can we fix it for the other tests too please?  It should be pretty trivial unless the helpers are gone.

src/bootstrap_handler.rs
Fraser999: Sorry - I meant `assert!(BootstrapHandler::new().unwrap().read_bootstrap_file().is_ok());`

src/bootstrap_handler.rs
mmoadeli: @Fraser999 the helper does not exist. It has to be written

src/bootstrap_handler.rs
mmoadeli: Can they be addressed together?

mmoadeli:
@Fraser999  I have addressed all issues but unwrap and  random addresses.

mmoadeli:
@Fraser999  there is already an issue for random endpoint in crust

src/bootstrap_handler.rs
Fraser999: That's disappointing.  Maybe I'm remembering back to the C++ code.  I guess implementing such helpers would be beyond the scope of this task.  I'll add an issue for it.

Fraser999: There's already an open issue for this: https://github.com/maidsafe/crust/issues/120.

src/bootstrap_handler.rs
Fraser999: I'm not sure.  Obviously the best outcome is to have no `unwrap`s left in the production code, but at the very least this PR shouldn't be adding more!  If it's quick to just remove the pre-existing ones I'd probably just do them too.  Probably best to just clear your decision with @Viv-Rajkumar.

Fraser999:
The `unwrap`s are a blocker as far as I'm concerned.  There are 17 new instances of these introduced by this PR from what I can see.

src/bootstrap_handler.rs
dirvine: Here is something that may help `use net::test::{next_test_ip4, next_test_ip6};`

src/bootstrap_handler.rs
dirvine: I would say absolutely, lets not introduce any unwraps (Except for tests) and remove existing ones as well.

mmoadeli:
@Fraser999 
1- the println! are there to be able visually see what config has.  That's the way I do manual test
2- I remove the unwraps introduced by this PR as soon as I can.

Fraser999:
@mmoadeli 
> 1- the println! are there to be able to see what config has. That's the way I do manual test

They normally shouldn't remain beyond debugging though - they shouldn't be part of a PR like this.  If you're not using this to debug but instead to check that the contents of the file are correct, then whatever it is you're looking for visually should probably be added as a test?  Also, David posted the following in the Rust Slack channel:

> So to be clear, in libraries now we have effectively banned println! and will use env_logger for now.

mmoadeli:
@Fraser999  I remove the read tests then. Only write will stay there.

src/bootstrap_handler.rs
Fraser999: @dirvine Unfortunately those functions always return the loopback address, only the port changes.  Also, it looks like they're marked "experimental".

PR #255, opened 2015-08-18
use port from CLI in example

benjaminbollen:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/255)


maidsafe-highfive:
r? @vinipsmaker

PR #258, opened 2015-08-19
WIP Fixes #248 - Peer gets many NewConnections

### LINKED ISSUE #248: Peer gets many NewConnections , seemly not connected to another peer
this relates to, or is very similar to the routing issue raised https://github.com/maidsafe/routing/issues/535

It seems that the second (not the first), third, ... peer can occasionally get a flood of new connections on a very regular interval (every 0,5 ~ 1 sec), and they directly disconnect afterwards.

This seems to stop after more nodes join the network.

I'll post routing example output as soon as it re-occurs

@rossmuir also observed similar behaviour in the routing example?

chandraprakash:
Fix is in place now. Working on adding a test to replicate actions performed by routing.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/258)

### DIFF
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -295,7 +295,7 @@ impl ConnectionManager {
                         break;
                     },
                     Err(_) => {
-                        // debug!("Failed to get at least one bootstrap connection. continuing bootstrap loop");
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @s-brian

inetic:
Fixed by other PR

PR #270, opened 2015-08-27
Fixes sample config file and adds test to check it is correct

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/270)

### DIFF
--- a/src/config_handler.rs
+++ b/src/config_handler.rs
@@ -123,4 +123,40 @@ mod test {
             Err(_) => (),
         };
     }
+
+    #[test]
+    fn parse_sample_config_file() {
+        use ::std::path::Path;
+        use std::io::Read;
+        use ::super::Config;
+        use ::rustc_serialize::json;
+
+        let path = Path::new("installer/sample.config").to_path_buf();
+
+        let mut file = match ::std::fs::File::open(path) {
+            Ok(file) => file,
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @vinipsmaker

src/config_handler.rs
```diff
@@ -123,4 +123,40 @@ mod test {
...
+            Err(what) => {
+                assert!(false, format!("Error opening sample.config: {:?}", what));
+                return;
+            }
+        };
+
+        let mut encoded_contents = String::new();
+
+        match file.read_to_string(&mut encoded_contents) {
+            Ok(_) => {;},
```
vinipsmaker: `()` or `{}` instead weird `{;}`?

src/config_handler.rs
```diff
@@ -123,4 +123,40 @@ mod test {
...
+            Err(what) => {
+                assert!(false, format!("Error reading sample.config: {:?}", what));
```
vinipsmaker: use `panic!`? seems more appropriate here (and will free you from the `return`

vinipsmaker:
I forgot to use reviewable to do the review, sorry.

inetic:
Thanks, fixed.

inetic:
Np. Fixed

PR #275, opened 2015-08-28
Reading config files now terminates any dependent application if the …

Fraser999:
…file is unreadable for any reason other than non-existence.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/275)


maidsafe-highfive:
r? @inetic

PR #279, opened 2015-08-28
This should stop the infinite bootstrap loop

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/279)


maidsafe-highfive:
r? @vinipsmaker

dirvine:
Nice one @inetic

PR #291, opened 2015-08-31
Documentation updates.

Fraser999:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/291)


maidsafe-highfive:
r? @vinipsmaker

vinipsmaker:
Where is `../file_handler/struct.FileHandler.html`?

Fraser999:
It should be the path to http://maidsafe.net/crust/master/crust/file_handler/struct.FileHandler.html relative to http://maidsafe.net/crust/master/crust/connection_manager/struct.ConnectionManager.html which is where this comment ends up being generated.  I hope :)  The link works when I generate the docs locally with `cargo doc --no-deps`.

PR #293, opened 2015-08-31
initial messaging layer

vinipsmaker:
An alternative approach could be:

```
enum Event {
    Contacts(Contacts),
}

enum Message {
    External(Bytes),
    Internal(Event),
}
```

But this would be 2 levels of indirection that could be leading to
more inefficient representation. Also, current approach saves typing.

This commits also gets rid of parse_contacts, as the serialization
happens on a central placce.

This support is "initial" because there room for improvement, like adding other
events (e.g. ping...) and find code that could be obsoleted by this approach.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/293)

### DIFF
--- a/src/beacon.rs
+++ b/src/beacon.rs
@@ -266,6 +266,7 @@ mod test {
     use super::*;
     use std::thread;
     use transport;
+    use transport::Message;
 
     #[test]
     fn test_beacon() {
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @inetic

PR #297, opened 2015-09-01
Reverted the API change for `ConnectionManager::new()` so that it doe…

Fraser999:
…sn't return a `Result`.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/297)


maidsafe-highfive:
r? @vinipsmaker

PR #299, opened 2015-09-01
Fixes compilaton of the benchmark

inetic:
Though it fails to run, I believe it is because UDP beacon packets are being received from previous runs.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/299)

### DIFF
--- a/benches/connection_manager.rs
+++ b/benches/connection_manager.rs
@@ -24,9 +24,7 @@ use rand::random;
 use test::Bencher;
 use crust::*;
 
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @vinipsmaker

vinipsmaker:
It surely compiles now.

But still not quite there...

    test send_random_data ... thread '<main>' panicked at 'Unexpected event: NewConnection', benches/connection_manager.rs:89

PR #314, opened 2015-09-10
Channels

inetic:
Mutex locking is now gone in favor of channel approach. Two bootstrapping tests are commented out because bootstrapping refactor shall be my next task.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/314)


maidsafe-highfive:
r? @vinipsmaker

vinipsmaker:
That was a big one.

PR #324, opened 2015-09-16
fix #273

### LINKED ISSUE #273: crust_peer example only accepts TCP endpoints
The example currently interprets all addresses entered via the CLI as TCP.  It should allow connecting or sending a message to a TCP or UTP endpoint.  Just now, the example's input prompt is:

    Enter command (stop | connect <endpoint> | send <endpoint> <message>)>

 I can see a few options:

1. Have two separate "connect" commands and two separate "send" commands, e.g. `send_tcp <endpoint> <message> | send_utp <endpoint> <message>`.  This seems like the worst option to me, but is probably the quickest to implement.
2. Given an endpoint, try connect/send for each protocol and only fail if they both fail.  This should be fairly simple to implement too, and has the benefit of not negatively affecting the current user-experience.
3. Change the way we parse endpoints.  At the moment we parse something of the form `192.168.0.9:60231` or `[fe80::704b:62fc:5966:3b74]:60219` and otherwise fail.  However, we output endpoints to screen as `Utp(V4(192.168.0.9:60231))` or `Tcp(V6([fe80::704b:62fc:5966:3b74]:60219))`.  It would be easiest for users if we can parse what is output, since it's less confusing and they can then just double-click in the console to copy an endpoint - at least on Windows.
4. The most flexible would be to allow parsing `192.168.0.9:60231` *or* `Utp(V4(192.168.0.9:60231))` formats.  If the former, we handle that endpoint as described in point 2.  This would be the most work to implement (still not too much I think) but would give the best user-experience.

vinipsmaker:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/324)

### DIFF
--- a/examples/crust_peer.rs
+++ b/examples/crust_peer.rs
@@ -107,20 +107,29 @@ struct CliArgs {
 
 #[derive(Debug)]
 struct PeerEndpoint {
-    pub addr: SocketAddr,
+    pub addr: Endpoint,
 }
 
 impl Decodable for PeerEndpoint {
     fn decode<D: Decoder>(decoder: &mut D)->Result<PeerEndpoint, D::Error> {
         let str = try!(decoder.read_str());
-        let address = match SocketAddr::from_str(&str) {
+        if !str.ends_with(')') {
+            return Err(decoder.error("Protocol missing"))
+        }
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @inetic

PR #326, opened 2015-09-17
Adds test to show TcpStream has problem with concurency

inetic:
It can't be simultaneously written to and read from

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/326)


maidsafe-highfive:
r? @vinipsmaker

PR #328, opened 2015-09-17
Fixes the way how cbor::Decoder is used

inetic:
This fixes the 'network' test too.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/328)


maidsafe-highfive:
r? @vinipsmaker

vinipsmaker:
Good one :)

PR #330, opened 2015-09-18
add handshake to connection establishment

vinipsmaker:
Solves https://maidsafe.atlassian.net/browse/MAID-1353

Currently an "empty" message is sent as the handshake, but the plan is
to change the message to exchange public endpoints which will help with
hole punching.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/330)

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -28,7 +28,7 @@ use bootstrap_handler::BootstrapHandler;
 use config_handler::Config;
 use getifaddrs::{getifaddrs, filter_loopback};
 use transport;
-use transport::{Endpoint, Port, Message};
+use transport::{Endpoint, Port, Message, Handshake};
 use std::thread::JoinHandle;
 use std::net::{IpAddr, Ipv4Addr};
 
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @inetic

vinipsmaker:
- `service::test::bootstrap` test is failing because `Event::BootstrapFinished` is being emitted before `Event::OnConnect` and `Event::OnAccept`. Needs to delay new node confirmation until the handshake is finished.
- I'm still investigating why `service::test::connection_manager_start` is hanging. Hopefully is the same issue as above, then I can tackle the two problems with a single patch.

vinipsmaker:
`service::test::connection_manager_start` was hanging because a race in the test code itself. I'll fix the bootstrap and see if all tests which depend upon bootstrap are also fixed.

vinipsmaker:
hey @inetic, PR is ready for review (just waiting for CI tests).

vinipsmaker:
All tests are passing and I'm seeing a green light from the CI machines.

src/transport.rs
```diff
@@ -180,6 +180,11 @@ impl Ord for Endpoint {
 
 #[derive(Debug, Clone, RustcDecodable, RustcEncodable)]
 pub enum Message {
+    /// After a connection is established, peers should exchange a handshake.
+    ///
+    /// Only after the handshake is exchanged, crust should generate new
+    /// connection events.
+    Handshake,
```
inetic: I think `Handshake` should not be part of the `Message` enum as we'll never send `UserBlob` nor `Contacts` while handshaking and we'll never send `Handshake` while receiving `UserBlob` or `Contacts`.

src/state.rs
```diff
@@ -418,12 +439,9 @@ mod test {
...
         let accept_thread = thread::spawn(move || {
-            // Currently acceptor starts accepting implicitly,
-            // once fixed (https://github.com/maidsafe/crust/issues/316),
-            // the following lines should be uncommented.
-            //for a in acceptors {
-            //    assert!(::transport::accept(&a).is_ok());
-            //}
+            for a in acceptors {
+                assert!(State::accept(&a).is_ok())
+            }
```
inetic: I'm not sure this has changed, the acceptor still seems to start accepting implicitly on creation:

https://github.com/maidsafe/crust/blob/d9aa489cc4ef84d0557c983c50650334cbd52141/src/transport.rs#L308
https://github.com/maidsafe/crust/blob/d9aa489cc4ef84d0557c983c50650334cbd52141/src/tcp_connections.rs#L40

vinipsmaker: Yes, but now it's required for us to left these lines uncommented so we can initiate the handshake. Therefor, leaving the comment would be misleading (uncommenting code that is already uncommented).

inetic: I see, thanks

src/transport.rs
vinipsmaker: Solved. Just waiting for the CI tests.

PR #335, opened 2015-09-21
Do not panic if sending fails 

inetic:
Also minor documentation fixes

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/335)

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -227,12 +227,13 @@ impl State {
         });
     }
 
-    fn unregister_connection(&mut self, connection: Connection) {
-        if self.connections.remove(&connection).is_some() {
-            // Only send the event if the connection was there
-            // to avoid duplicate events.
-            let _ = self.event_sender.send(Event::LostConnection(connection));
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @vinipsmaker

src/state.rs
```diff
@@ -227,12 +227,14 @@ impl State {
         });
     }
 
-    fn unregister_connection(&mut self, connection: Connection) {
-        if self.connections.remove(&connection).is_some() {
+    pub fn unregister_connection(&mut self, connection: Connection) {
+        if self.connections.remove(&connection).is_none() {
             // Only send the event if the connection was there
```
vinipsmaker: I like the approach to return early if some condition prevent the logic. I saw Linus Torvalds defending this style too (a shame I didn't save the link).

Anyway... this comment looks half out of place once you moved the notification code. Maybe something like `// Avoid duplicate connection cleanup`?

vinipsmaker:
Looks good, but waiting for CI just in case.

PR #339, opened 2015-09-22
Fixes and simplifies the 'network' test

inetic:
Was only panicking on CI machines, and even then the test
resulted with OK for some reason.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/339)

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -385,11 +385,10 @@ mod test {
     use connection::Connection;
     use std::thread::spawn;
     use std::thread;
-    use std::sync::mpsc::{Receiver, Sender, channel};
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @vinipsmaker

src/service.rs
```diff
@@ -585,155 +542,96 @@ mod test {
 
     #[test]
     fn network() {
+        //extern crate env_logger;
```
vinipsmaker: remove the commented code?

vinipsmaker:
All else seems fine.

src/service.rs
inetic: Done

PR #348, opened 2015-10-04
In tests, do not connect to unspecified addrs

inetic:
Windows doesn't seem to handle it

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/348)


maidsafe-highfive:
r? @vinipsmaker

Issue #327, opened 2015-09-17
crust_peer fails to stop


vinipsmaker:
Not the first time it happens ( https://github.com/maidsafe/crust/issues/277 https://github.com/maidsafe/crust/issues/141 ).

vinipsmaker:
It's working again. Not sure which commit fixed it, but must have been work from @inetic.

vinipsmaker:
It's failing again.

## How to reproduce

1. Open a **single** `crust_peer` instance.
2. Do not connect to any other node. You may see:

           Running `target/debug/examples/crust_peer`
           Listening on ports Tcp(V4(0.0.0.0:5483))
           Enter command (stop | connect <endpoint> | send <endpoint> <message>)>

  You might also see:

      Didn't bootstrap to an existing network - this may be the first node of a new network.
3. Enter `stop`.
4. `crust_peer` doesn't stop.

A hint to solve this issue. The hanged/hanging crust_peer instance will gracefully stop if you, now, start another `crust_peer` instance that will automatically connect (through beacon) to the previous instance. This second instance won't have any trouble execute the `stop` command.

inetic:
Fixed here https://github.com/maidsafe/crust/commit/20965f1f6eeaf178572964dba854f97058730677

PR #351, opened 2015-10-07
Enter REPL without needing to bootstrap first.

canndrew:
Fix for bug #327. With this, only the `running_speed_test` executing branch will block waiting for a bootstrap connection.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/351)

### DIFF
--- a/examples/crust_peer.rs
+++ b/examples/crust_peer.rs
@@ -21,10 +21,11 @@
 #![deny(bad_style, deprecated, drop_with_repr_extern, improper_ctypes, non_shorthand_field_patterns,
         overflowing_literals, plugin_as_library, private_no_mangle_fns, private_no_mangle_statics,
         raw_pointer_derive, stable_features, unconditional_recursion, unknown_lints,
-        unsafe_code, unused_allocation, unused_attributes,
+        unused_allocation, unused_attributes,
         unused_comparisons, unused_features, unused_parens, while_true)]
 #![warn(trivial_casts, trivial_numeric_casts, unused, unused_extern_crates, unused_import_braces,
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
Thanks for the pull request, and welcome! The MaidSafe team is excited to review your changes, and you should hear from @inetic (or someone else) soon.

vinipsmaker:
Please re-enable `unsafe_code` lint

canndrew:
That would mean forgoing the use of `select!`. I think we need to decide on a general policy regarding whether we want to allow `select!` or not. It's pretty much a necessity in code that uses lots of channels.

Fraser999:
Our current policy is to try and get off the Nightly channel, so effectively this precludes the use of `select!`.  I haven't any direct experience of using it, but if it's as helpful as you're indicating, should we be looking at implementing our own copy until the official one is stabilised?

Also, we're using a standard set of lint settings across all of our projects now, and I think any changes to these should be discussed before being made.  They were added in a rush, so I expect some or all could be debated, but it would be good as a company to get into a habit of not changing these without prior discussion and agreement.

canndrew:
> but if it's as helpful as you're indicating, should we be looking at implementing our own copy until the official one is stabilised?

[Rust's tracking issue](https://github.com/rust-lang/rust/issues/27800) for stabilizing `select!` lists some third-party alternatives but I haven't looked into any of them.

> Also, we're using a standard set of lint settings across all of our projects now, and I think any changes to these should be discussed before being made. They were added in a rush, so I expect some or all could be debated, but it would be good as a company to get into a habit of not changing these without prior discussion and agreement.

Agreed.

PR #356, opened 2015-10-12
Fixes compilation of the simple_sender example

inetic:
Compilation failed due to new compiler I believe

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/356)


maidsafe-highfive:
r? @vinipsmaker

PR #352, opened 2015-10-07
Updated lint checks and README.md

maqi:
replacing PR https://github.com/maidsafe/crust/pull/349

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/352)

PR #360, opened 2015-10-14
MAID-1458 Add new example: reporter.rs

madadam:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/360)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,3 +44,9 @@ name = "crust_peer"
 path = "examples/crust_peer.rs"
 test = true
 bench = false
+
+[[example]]
+name = "reporter"
+path = "examples/reporter.rs"
+test = true
+bench = false
--- a/examples/reporter.rs
+++ b/examples/reporter.rs
@@ -1,0 +1,305 @@
+// Copyright 2015 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under (1) the MaidSafe.net Commercial License,
+// version 1.0 or later, or (2) The General Public License (GPL), version 3, depending on which
+// licence you accepted on initial access to the Software (the "Licences").
+//
+// By contributing code to the SAFE Network Software, or to this project generally, you agree to be
+// bound by the terms of the MaidSafe Contributor Agreement, version 1.0.  This, along with the
+// Licenses can be found in the root directory of this project at LICENSE, COPYING and CONTRIBUTOR.
+//
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+//! Example for testing direct connections between nodes.
+
+#![forbid(missing_docs, warnings)]
+#![deny(bad_style, deprecated, drop_with_repr_extern, improper_ctypes, non_shorthand_field_patterns,
+        overflowing_literals, plugin_as_library, private_no_mangle_fns, private_no_mangle_statics,
+        raw_pointer_derive, stable_features, unconditional_recursion, unknown_lints,
+        unsafe_code, unused_allocation, unused_attributes,
+        unused_comparisons, unused_features, unused_parens, while_true)]
+#![warn(trivial_casts, trivial_numeric_casts, unused, unused_extern_crates, unused_import_braces,
+        unused_qualifications, unused_results, variant_size_differences)]
+
+extern crate crust;
+extern crate docopt;
+extern crate env_logger;
+
+#[macro_use]
+extern crate log;
+
+extern crate rand;
+extern crate rustc_serialize;
+extern crate time;
+
+use crust::{Endpoint, Event, FileHandler,  Port, Service};
+use docopt::Docopt;
+use rand::{thread_rng, Rng};
+use rustc_serialize::{Decodable, Decoder, Encodable, Encoder};
+use std::collections::HashMap;
+use std::io;
+use std::path::Path;
+use std::sync::mpsc::channel;
+use std::thread;
+use time::Tm;
+
+static USAGE: &'static str = "
+Usage:
+    reporter <config>
+";
+
+const MIN_RUN_TIME_MS: u32 = 1000;
+const MAX_RUN_TIME_MS: u32 = 2500;
+
+fn main() {
+    match env_logger::init() {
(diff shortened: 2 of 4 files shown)


maidsafe-highfive:
Thanks for the pull request, and welcome! The MaidSafe team is excited to review your changes, and you should hear from @vinipsmaker (or someone else) soon.

Fraser999:
This is almost good to merge, but I have a few general comments (roughly in descending order of importance):

- There's no documentation on how to use this example.  We should probably have this available by running the example with `--help`.  It should provide the details as listed in the original JIRA task and subsequent comments and also mention the example input files in the reporter subdir.  This is the only blocker to merging this PR.
- Because there is no throttling of the rate of sending messages, and because we log each individual event, the logfiles will be unreasonably large for the scenario in which we'd intended to use this example.  I tried with just three nodes, each connected to the other two, each with `"service_runs": 10,` and `"max_wait_before_restart_service_secs": 1`.  This resulted in logfiles of around 30MB each (~600,000 lines).  We're planning to use this example to interconnect 64 nodes, probably for durations of at least several minutes, possibly hours.  The JIRA task didn't specify these requirements (although it did mention "a few messages"), so I think it's reasonable to merge this PR and address this in a separate commit.  We'll need to specify how we want to handle the issue before tackling it.  I'll add a GutHub issue for this.
- I think Crust is the only MaidSafe project which still hasn't adopted the "official" set of lint checks.  https://github.com/maidsafe/crust/pull/352/files#diff-b4aea3e418ccdb71239b96952d9cddb6 fixes this for the main library, but it would be good to use the set of checks defined in that PR.
- We don't have an official style guide yet, but some things which have been agreed are
    - avoid `use` statements as far as possible (if they're required to bring a trait into scope, limit the `use` to the affected function)
    - line length <= 100 chars
    - use meaningful variable names - don't abbreviate words unless the abbreviation is well known.  In particular avoid single char variable names except e.g. maybe in for loops

So, in summary, if you could fix the documentation issue, I'll merge this.  The lint checks I think can be fixed without changing any other code, so it'd be good to do that too if that's the case.  The logfile size issue and the style issues can be left unchanged and will be handled separately.


---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/360)

inetic:
@Fraser999 nice review. But I strongly disagree about the "avoid `use` statements as far as possible" style guide rule. I spoke with Spandan the other day (as he wrote the rules) and he wasn't too strong about this neither. While I would see why we shouldn't use `use foo::*` or `use foo::bar; // if we also use bar from other module`, or if it's a name that is used only a few times in the code, but this rule applied generically to any case just adds noise to the code and makes it harder to read.

madadam:
Thanks for the review @Fraser999 . I added the documentation. Also agree with the throttling, this should definitely be addressed at some point. 

About the `use` rule: I'm with @inetic on this. Don't understand what's wrong in using `use`. Perhaps care to elaborate?

ustulation:
When i code a library i immediately know all types i see and what they are, but when i see someone else's it becomes a horrible experience if i have to endlessly keep bouncing up and down the page to know what that type is or where it originated from, whether it is standard type or user defined type etc. It takes infinitely more time to understand what has been done. If it is one file (like main) that is aggregating everything eventually then it's still ok, but otherwise i find it frustrating to look at a huge library with every file having hundreds of `use` qualifiers at the top. In  absence of proper IDE (on editors reading code is like reading code from github) i can't ctrl-click or something to go quickly to the source which would have salvaged the situation quite a bit.

But having said all that, i realise that it is just a preference thing and how one's brain is oriented (or disoriented). So i wasn't too strong on this in discussion with @inetic . I however do use it in crates i am involved in and has helped reviewing and understanding of code much faster.

inetic:
> but when i see someone else's it becomes a horrible experience if i have to endlessly keep bouncing up and down the page to know what that type is or where it originated from

Must say I've never felt this discomfort.

>  whether it is standard type or user defined type,

Even for types such as BTreeMap, BTreeSet, Arc, ...? These are so ubiquitous that everyone probably knows where they originated. And even if not, knowing the exact module location for these types is hardly ever useful, and for the times when it is one can just "bounce up and down" (not endlessly though).

> every file having hundreds of use directives at the top.

This is not the case in crust I believe.

> But having said all that, i realise that it is just a preference thing

Yes, I think this is the key, but I find forcing preference from one library to all the others unfair, especially if it inconveniences others. Also, looking around other projects, it appears to me that this in not a widely adopted preference.

As said, there are cases like the three I listed above where it makes sense. But in most cases this rule clutters the code and obfuscates the logic. Take for example this function from this PR:

```
    fn ensure_port(&mut self, port: u16) {
        use std::net::SocketAddr::{V4, V6};
        use std::net::{SocketAddrV4, SocketAddrV6};

        if self.0.port() != 0 { return; }

        self.0 = match self.0 {
            V4(addr) => V4(SocketAddrV4::new(*addr.ip(), port)),
            V6(addr) => V6(SocketAddrV6::new(*addr.ip(), port, addr.flowinfo(), addr.scope_id()))
        }
    }
```

With this rule it would be:

```
    fn ensure_port(&mut self, port: u16) {
        if self.0.port() != 0 { return; }

        self.0 = match self.0 {
            ::std::net::SocketAddr::V4(addr) => ::std::net::SocketAddr::V4(::std::net::SocketAddrV4::new(*addr.ip(), port)),
            ::std::net::SocketAddr::V6(addr) => ::std::net::SocketAddr::V6(::std::net::SocketAddrV6::new(*addr.ip(), port, addr.flowinfo(), addr.scope_id()))
        }
    }
```

Which, for one, goes against the "max 100 chars per line" rule, for two (and I don't think this is subjective) makes it harder to read by adding unnecessary noise.

Fraser999:
@inetic and @madadam Yeah - I think @ustulation (who initially came up with the rule I think) has nicely summarised the rationale.  I have to say that I'm generally in the same position as him on this point, but I don't think this is the place to discuss it.  There's an RFC planned to discuss and resolve these issues, but until then it was stated on Slack that we should all follow this guideline.  I don't think it means rewriting existing code (yet), but rather that all new code should adhere to our rules.  If it needs further discussion immediately, I'd suggest moving the conversation to Slack or email.

@madadam Thanks for the update - a nice, clean PR.  If you could provide us with your bitcoin wallet address, we'll get you paid.

Fraser999:
I have filed https://github.com/maidsafe/crust/issues/364 to address the logfile size problem.

madadam:
1J4t5y4XJisLPv7tvCJcpho9GweM9LRoEx. Thanks.

NickLambert:
Payment is on the way, many thanks for your help!

PR #376, opened 2015-10-23
Join threads, release bound sockets on Service::stop 

### LINKED ISSUE #361: Service keeps holding the accepting UTP port even after stopped
Steps to reproduce:

1. Create new Service: `let service = Service::new()`
2. Call start_accepting with a UTP port on it: `service.start_accepting(Port::Utp(1234))`
3. Stop it: `service.stop()`
4. Repeat 1 and 2, using the same port.

The code panics with "Address already in use" error. This only happens when accepting on UTP ports, not TCP. My guess is that there is a thread somewhere which handles the accepting and this thread is not properly disposed of when the service is stopped.

This issue affects https://maidsafe.atlassian.net/browse/MAID-1458. Until it is fixed, the example can work only with TCP ports.

madadam:
1. Adds join of the `State` thread on `Service::stop`
2. Simplifies acceptor logic (both TCP and UTP) by removing unnecessary threads. This also fixes #361 .

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/376)

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -50,6 +50,7 @@ pub struct Service {
     beacon_guid_and_port : Option<(beacon::GUID, u16)>,
     config               : Config,
     cmd_sender           : Sender<Closure>,
+    state_thread_handle  : Option<JoinHandle<()>>,
 }
 
 impl Service {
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @vinipsmaker

src/service.rs
```diff
@@ -299,6 +312,8 @@ impl Service {
                 let cmd_sender3 = cmd_sender2.clone();
 
                 let _ = cmd_sender2.send(Box::new(move |state: &mut State| {
+                    state.listening_ports.remove(&port);
```
vinipsmaker: I'm having trouble to understand this line. Care to elaborate?

madadam: Sure. Every time the service is about to accept a new connection, the port the acceptor listens on is stored in this hashset (see line 304). Then once something is accepted, the port is removed. So the hashset keeps tracks of the ports the acceptors are currently holding while they are blocked waiting for incoming connection. This hashset is then used in the `stop` method (lines 227 to 234) where dummy connects are made to these ports, to "wake up" the blocked acceptors, so they can be destroyed.

Issue #274, opened 2015-08-27
Define exit codes


Fraser999:
The `FileHandler` class will cause any dependent binaries to terminate if it can't parse a given file.  We need to choose and document the value to use for the exit codes.

Currently I'm using the following:

| Reason | Exit Code |
|:---|:---:|
| Failed to open for any reason other than non-existence | 1 |
| Failed to read | 2 |
| Failed to decode as JSON | 3 |

madadam:
I believe this can be closed as `ConfigFileHandler` (or any other library we depend on) no longer calls `exit`.

PR #380, opened 2015-10-28
Fix file handling

canndrew:
This PR:
* gets rid of the `abort`s in file handling code
* gets rid of the meaningless `NotSet` error variant
* fixes/simplifies a test

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/380)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -28,8 +28,12 @@ pub enum Error {
     JsonEncoderError(::rustc_serialize::json::EncoderError),
     /// Wrapper for a `::cbor::CborError`
     CborError(::cbor::CborError),
-    /// Indicates variable has never been set
-    NotSet,
+    /// The file was found. Either we don't know where to look for it or no file existed at any of
+    /// the candidate locations.
+    ReadFileFailed,
+    /// Failed to write a file. Either we don't know where to write it to or writing to all the
+    /// candidate locations failed.
+    WriteFileFailed,
 }
 
 impl From<::std::env::VarError> for Error {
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @vinipsmaker

Fraser999:
The aborts were by design and were debated fairly extensively at the time.  Has their removal been approved by @dirvine ?

If so, we should also ask Scott to update the flowchart (https://github.com/maidsafe/crust/blob/master/docs/vault_config_file_flowchart.pdf) to rename the "Terminate" blocks.

vinipsmaker:
Probably related #274

canndrew:
Is there a link to that debate somewhere? I can't see why it would ever make sense to abort from a library. If the library consumer wants to abort when they get an `Err` then they can choose to do that.

Fraser999:
> Is there a link to that debate somewhere?

Not that I know of.  I remember it mostly being discussed on Slack and Mumble.

dirvine:
I think it's fine to pass these up as long as they are not swallowed by upper layers as basically the abort here means crust will not function and can do nothing. So these are errors that defo cannot be ignored. I am more inclined to start saying at this stage no error should be swallowed, even IO errors out of our control. Perhaps anywhere we feel the need to swallow an error we should have a warning message. 
Anyway that's a bigger discussion, passing these up is OK as long as they are definitely not ignored as crust is dead, so if an upper layer can rewrite the config to work then great but at the moment they know nothing of crust config and components (crusts job is to get you back on a network and keep you there, upper layers do not even know IP addresses etc. never mind seed nodes and the like).

vinipsmaker:
> This branch has conflicts that must be resolved

PR #386, opened 2015-10-29
Fix: crust_peer showing invalid connection id after reconnection

### LINKED ISSUE #383: crust_peer example fails to keep list of peers synchronised
When the example loses a connection it appears that this is not properly handled.

This is exposed when a subsequent connection is established and we try to send a message to the new connection.  The index shown for the new connection (via the `list` command) appears to be correct, but using the `send` command only succeeds if we send to the index + no. of dropped connections.

inetic:
fixes #383

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/386)

### DIFF
--- a/examples/crust_peer.rs
+++ b/examples/crust_peer.rs
@@ -247,21 +247,17 @@ impl Network {
     }
 
     pub fn print_connected_nodes(&self) {
-        let connected_nodes =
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @canndrew

PR #389, opened 2015-10-30
Duration precision fix

inetic:
When the timeout was less than 1ms but bugger than 0ms, the
conversion to ms and then back to Duration resulted in
zero duration, which the set_read_timeout fn did not like
on windows.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/389)


maidsafe-highfive:
r? @vinipsmaker

Issue #393, opened 2015-11-02
AppVeyor CI timing out on windows


inetic:
Seems to have started with [this PR](https://github.com/maidsafe/crust/pull/376) (the CI machined did not report it as failure at the time).

inetic:
Update: Fraser seems to be able to reproduce this if he enables two network interfaces in his windows PC:

>  I was just running the exact commands which the CI machines run to see if I could replicate the hanging on Windows. I can make it hang every time if I enable 2 network cards on my machine.  If I only enable 1 card, the tests always exit cleanly.

inetic:
It also only happens if all the tests are executed (`cargo test`), if one executes the tests one by one, they all succeed.

```
cargo test --release -- --nocapture beacon::test::test_avoid_beacon
cargo test --release -- --nocapture beacon::test::test_beacon
cargo test --release -- --nocapture bootstrap_handler::test::duplicates
cargo test --release -- --nocapture bootstrap_handler::test::max_contacts
cargo test --release -- --nocapture bootstrap_handler::test::prune
cargo test --release -- --nocapture bootstrap_handler::test::serialise_and_parse
cargo test --release -- --nocapture config_handler::test::parse_sample_config_file
cargo test --release -- --nocapture config_handler::test::read_config_file_test
cargo test --release -- --nocapture file_handler::test::read_write_file_test
cargo test --release -- --nocapture getifaddrs::test::test_filter_loopback
cargo test --release -- --nocapture getifaddrs::test::test_getifaddrs
cargo test --release -- --nocapture hole_punching::tests::test_get_mapped_socket_from_self
cargo test --release -- --nocapture hole_punching::tests::test_udp_hole_punching_0_time_out
cargo test --release -- --nocapture hole_punching::tests::test_udp_hole_punching_1_terminates_no_secret
cargo test --release -- --nocapture hole_punching::tests::test_udp_hole_punching_2_terminates_with_secret
cargo test --release -- --nocapture map_external_port::test::upnp ... ignored
cargo test --release -- --nocapture sequence_number::test::sequence_number
cargo test --release -- --nocapture service::test::bootstrap
cargo test --release -- --nocapture service::test::connection_manager
cargo test --release -- --nocapture service::test::connection_manager_start
cargo test --release -- --nocapture service::test::network
cargo test --release -- --nocapture service::test::reaccept
cargo test --release -- --nocapture service::test::rendezvous_connection
cargo test --release -- --nocapture state::test::bootstrap_off_list
[...]

PR #400, opened 2015-11-04
Win freeze

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/400)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ license = "GPL-3.0"
 homepage = "http://maidsafe.net"
 
 [dependencies]
-libc = "*"
+libc = "0.1.12"
 cbor = "*"
 docopt = "*"
 rand = "*"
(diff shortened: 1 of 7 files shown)


maidsafe-highfive:
r? @canndrew

inetic:
fixes #393

vinipsmaker:
So the problem was in the test and not on our code?

inetic:
Depends on what you mean by "our code" :), but yes, the thread in that test never finished. My theory is that we've had this bug there a long time but perhaps something changed on the test machines which made them freeze.

Viv-Rajkumar:
:+1: nice one @inetic

canndrew:
Looks like these commits conflict a bit with the [fix file handling](https://github.com/maidsafe/crust/pull/380) PR.

inetic:
@canndrew yes, sorry about that, I needed to debug/print the location of the cache file and it was very hard to do with [this code](https://github.com/maidsafe/crust/pull/380/files#diff-dd722105e2a87507c27621ae03272916L78), it already bugged me for some time so I just changed it.

Issue #388, opened 2015-10-30
Possible self-bootstrapping


benjaminbollen:
As discovered by debugging the routing connection management an intermittent failure (perhaps related to the bootstrap cache file, but a unvalidated guess), occurs for crust 0.4.2, where a single node can bootstrap to itself.

@vinipsmaker 's test does not showcase this issue (so far), which might indicate some bootstrap cache file residue causing it?
```
 #[test]
    fn examine_self_bootstrapping() {
        let (tx, rx) = ::std::sync::mpsc::channel();
        let mut service = ::crust::Service::new(tx).unwrap();
        service.bootstrap();
        let timeout = ::time::Duration::seconds(5);
        let start = ::time::now();
        let mut result = Err(::std::sync::mpsc::TryRecvError::Empty);
        while ::time::now() < start + timeout && result.is_err() {
            result = rx.try_recv();
            if let Ok(ref ev) = result {
                match *ev {
                    ::crust::Event::OnAccept(ref ep) => {
                        panic!("What the hell? {:?}", ep);
                    },
                    _ => {
                        panic!("aaa {:?}", *ev);
                    },
                }
            }
            ::std::thread::sleep_ms(100);
        }
    }
```

benjaminbollen:
Yesterdays observations (repeated from issue https://github.com/maidsafe/routing/issues/757) were eg :

```
$ RUST_LOG=routing=debug ./target/release/examples/key_value_store --node
INFO:routing::routing_node: RoutingNode Client(e61f88..ed1119) listens on [Tcp(V4(127.0.0.1:5483)), Tcp(V4(192.168.0.3:5483))]
DEBUG:routing::routing: Started routing run().
DEBUG:routing::routing_node: RoutingNode started running and started bootstrap
DEBUG:routing::routing_node: handle_on_connect: Disconnected adding unknown bootstrap Connection(Tcp V4(127.0.0.1:55474) -> V4(127.0.0.1:5483))
DEBUG:routing::routing_core: add_unknown_bootstrap_connection: added Connection(Tcp V4(127.0.0.1:55474) -> V4(127.0.0.1:5483))
DEBUG:routing::routing_core: add_unknown_bootstrap_connection: set state Bootstrapped
DEBUG:routing::routing_node: Saying hello I am Client(e61f88..ed1119) on Connection(Tcp V4(127.0.0.1:55474) -> V4(127.0.0.1:5483)), confirming None
[...]

PR #402, opened 2015-11-04
do not adding any non-global addresses to cache anymore

vinipsmaker:
Updates #388

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/402)

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -373,6 +373,27 @@ pub fn random_endpoints(count: usize) -> Vec<::transport::Endpoint> {
 }
 
 #[cfg(test)]
+pub fn random_global_endpoint() -> ::transport::Endpoint {
+    // TODO - randomise V4/V6 and TCP/UTP
+    let address = ::std::net::SocketAddrV4::new(
+        ::std::net::Ipv4Addr::new(173, // ensure is a global addr
+                                  ::rand::random::<u8>(),
+                                  ::rand::random::<u8>(),
+                                  ::rand::random::<u8>()),
+        ::rand::random::<u16>());
+    ::transport::Endpoint::Tcp(::std::net::SocketAddr::V4(address))
+}
+
+#[cfg(test)]
+pub fn random_global_endpoints(count: usize) -> Vec<::transport::Endpoint> {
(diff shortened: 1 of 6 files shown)


maidsafe-highfive:
r? @inetic

vinipsmaker:
`service::test::bootstrap` fails after I filter out non-global addresses from the list of cached addresses.

After investigating the problem, here are my findings:

- The test start two `Service` objects, bootstraping both and hoping they'll find each other through beacon. Obvious.
- They DO find each other and a connection IS established.
- Different from normal connection establishment, the "connecting" side (function `state::seek_peers`) will actually drop the `Transport` object after receiving the first message, just to add the list of endpoints received in this first message to the list of endpoints to be attempted in the bootstrap attempt.
- By this time, the `BootstrapFinished` event will be already emitted.
- So, we've done a successful connection establishment with handshake and all neat stuff, just to change the state of `Service` to `BootstrapFinished` and then drop the "good" connection that we already had established.
- Beacon handling must be reworked.

vinipsmaker:
just in case the new build succeed by luck, here is the previous failing test: https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.374/job/a63lrqvf2rsho130#L892

vinipsmaker:
tests are passing, PR is ready

src/util.rs
```diff
@@ -355,7 +355,7 @@ pub fn loopback_if_unspecified(addr : IpAddr) -> IpAddr {
 pub fn random_endpoint() -> ::transport::Endpoint {
     // TODO - randomise V4/V6 and TCP/UTP
     let address = ::std::net::SocketAddrV4::new(
-        ::std::net::Ipv4Addr::new(::rand::random::<u8>(),
+        ::std::net::Ipv4Addr::new(173, // ensure is a global addr
                                   ::rand::random::<u8>(),
                                   ::rand::random::<u8>(),
                                   ::rand::random::<u8>()),
```
inetic: I think we should rename this function to `random_global_endpoint`.

vinipsmaker: Done

PR #409, opened 2015-11-09
Nightly fixes

canndrew:
Fixes for crust being broken on nightly.

**Note:** Commit a0b210e removes the `unused_results` lint because it apparently conflicts with `#[derive(Debug)]`. This commit should be reverted as soon as this is fixed upstream.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/409)


maidsafe-highfive:
r? @vinipsmaker

Issue #408, opened 2015-11-08
Cannot build crust on nightly


achanda:
Here is what I see on OSX and Ubuntu

```
Abhishek@dev:~/crust$ cargo test
    Updating registry `https://github.com/rust-lang/crates.io-index`
 Downloading hpack v0.2.0
 Downloading bitflags v0.3.2
 Downloading rand v0.3.11
 Downloading strsim v0.3.0
 Downloading num v0.1.27
 Downloading memchr v0.1.7
 Downloading igd v0.2.1
 Downloading time v0.1.33
 Downloading advapi32-sys v0.1.2
 Downloading temp_utp v0.6.17
 Downloading itertools v0.4.3
 Downloading num_cpus v0.2.7
 Downloading cookie v0.1.21
 Downloading bitflags v0.1.1
 Downloading env_logger v0.3.2
 Downloading byteorder v0.3.13
 Downloading unicase v1.0.1
 Downloading mime v0.1.0
 Downloading rustc-serialize v0.3.16
 Downloading httparse v1.0.0
 Downloading typeable v0.1.2
 Downloading nix v0.3.9
 Downloading traitobject v0.0.1
 Downloading libc v0.2.1
 Downloading term v0.2.12
 Downloading net2 v0.2.16
 Downloading docopt v0.6.74
 Downloading winapi v0.2.4
 Downloading crossbeam v0.1.6
 Downloading aho-corasick v0.3.4
 Downloading regex-syntax v0.2.2
 Downloading regex v0.1.41
 Downloading kernel32-sys v0.1.4
 Downloading hyper v0.6.15
 Downloading log v0.3.3
 Downloading solicit v0.4.4
 Downloading cbor v0.3.16
 Downloading xml-rs v0.1.26
 Downloading language-tags v0.0.7
 Downloading ws2_32-sys v0.1.0
 Downloading libc v0.1.12
 Downloading url v0.2.37
 Downloading matches v0.1.2
 Downloading cfg-if v0.1.0
 Downloading winapi-build v0.1.1
   Compiling byteorder v0.3.13
   Compiling winapi v0.2.4
   Compiling libc v0.2.1
   Compiling strsim v0.3.0
   Compiling bitflags v0.3.2
   Compiling traitobject v0.0.1
   Compiling itertools v0.4.3
   Compiling crossbeam v0.1.6
   Compiling libc v0.1.12
   Compiling language-tags v0.0.7
   Compiling typeable v0.1.2
[...]

PR #411, opened 2015-11-09
Nightly fixes

canndrew:
More changes for [this bug](https://github.com/maidsafe/crust/issues/408). This is an extension of [this PR](https://github.com/maidsafe/crust/pull/409).

**Note:** Commit 4a8b05c removes the `unused_results` lint because it apparently conflicts with `#[derive(Debug)]`. This commit should be reverted as soon as this is fixed upstream.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/411)


maidsafe-highfive:
r? @vinipsmaker

PR #414, opened 2015-11-11
adds a `Service` ctor that ignores beaconing

### LINKED ISSUE #413: Allow users to explicitly disable beaconing through the config file
As debated, it should work such that when a new config file is created, it would fill the beacon port to the default value (think it is 5483), but users should be able to explicitly set this value to `null` to disable beacon.

vinipsmaker:
Fixes #413.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/414)

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -70,6 +70,20 @@ impl Service {
         Service::construct(event_sender, config)
     }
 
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @inetic

PR #420, opened 2015-11-13
change API to notify connection failures

### LINKED ISSUE #418: `Event::OnConnect` should be fired even if connecting failed
The event [Event::OnConnect(Connection, u32 /* token */)](https://github.com/maidsafe/crust/blob/master/src/event.rs#L45) should be changed to `Event::OnConnect(io::Result<Connection>, u32 /* token */)` and it should be fired even if connection fails to be established (or if handshake fails).

vinipsmaker:
Fix #418

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/420)


maidsafe-highfive:
r? @inetic

Issue #419, opened 2015-11-12
Beacon should only be started explicitly using API, not the config file.


inetic:
The `Service` object should have a new public member function named `start_beacon(udp_port: u16) -> io::Result<u16>`, beacon should not be started implicitly through `Service::new` and any mention of beacon should be removed from the config file structure.

The tricky part about this task is that since beacon acceptor was previously created in the constructor, its presence was used as an indicator whether we can write into the bootstrap.cache file or not. Since we no longer can rely on it, we need other means to ensure our bootstrap.cache file doesn't get corrupted (written to by multiple processes).

Here is some brain dump of alternatives:

1. Use actual database
2. Use memory mapped files to sync file access between processes
3. Create a thread that will accept connections on a predefined TCP port, only one process per PC can successfully start such acceptor. Only the process with the acceptor can access the cache file, and all the other processes will need to communicate changes to the file through TCP connections.
4. All processes shall first write to a unique temporary files, once done, the temporary files will be renamed to the bootstrap.cache file.

dirvine:
:+1 for 2: as this seems the correct answer

vinipsmaker:
`1.` seems correct answer, but I didn't see a `SQLite` or alike for Rust. I also don't think it's correct to wait 5 years until somebody write one. `3.` is hacky and inconvenient. So I'd vote on `2.` .

vinipsmaker:
Previously we're only accepting the possibility to start beaconing once. Should we change that now?

vinipsmaker:
Solved in #422

PR #422, opened 2015-11-13
beaconing cleanup

vinipsmaker:
- Don't start beaconing implicitly anylonger.
- Eliminate options related to beaconing from config file.

API changes follow from these changes.

Still to do:

- `state::populate_bootstrap_contacts` should use another mechanism to
  detect availability of the bootstrap cache file.
- Now beacon isn't always used, so we cannot rely on the beacon port
  ownership to elect the process that should own the cache file.
- Should we allow beacon run on multiple ports? Maybe the answer should
  change now.

Update #419

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/422)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ net2 = "~0.2.17"
 temp_utp = "~0.6.19"
 log = "~0.3.3"
 crossbeam = "~0.1.6"
+memmap = "0.2.2"
 
 [dev-dependencies]
 env_logger = "~0.3.2"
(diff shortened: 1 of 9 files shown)


maidsafe-highfive:
r? @inetic

vinipsmaker:
The Windows failure is unrelated to our code and is already being worked on: https://github.com/seanmonstar/num_cpus/issues/15

Let's wait for the fix and retrigger the CI build.

vinipsmaker:
CI test is passing. PR can be merged now.

PR #429, opened 2015-11-20
Fix rust-utp usage.

vinipsmaker:
Rust's `std::io::Read` trait uses a 0-bytes read to signalize EOF[1].
rust-utp also follows this design.

[1] http://doc.rust-lang.org/nightly/std/io/trait.Read.html#tymethod.read

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/429)


maidsafe-highfive:
r? @inetic

PR #430, opened 2015-11-20
Remove lint and 100 character pernicketiness

rossmuir:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/430)


maidsafe-highfive:
r? @inetic

PR #434, opened 2015-11-26
Fix 433

vinipsmaker:
r? @canndrew

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/434)


vinipsmaker:
Tests passed.

PR #441, opened 2015-12-07
Use Endpoint instead of SocketAddr

canndrew:
In the OnConnect, OnRendezvousConnect and OnAccept events.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/441)


maidsafe-highfive:
r? @vinipsmaker

PR #442, opened 2015-12-08
Network test utp

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/442)

### DIFF
--- a/src/utp_wrapper.rs
+++ b/src/utp_wrapper.rs
@@ -27,7 +27,7 @@ impl UtpWrapper {
         socket.set_read_timeout(Some(CHECK_FOR_NEW_WRITES_INTERVAL_MS));
 
         let _ = thread::Builder::new().name("rust-utp multiplexer".to_string()).spawn(move || {
-            let mut socket_last_success = ::time::SteadyTime::now();        
+            let mut timeout = ::time::SteadyTime::now() + ::time::Duration::seconds(SOCKET_FAILURE_ALLOWANCE);
             'outer:
             loop {
                 let mut buf = [0; BUFFER_SIZE];
@@ -36,7 +36,7 @@ impl UtpWrapper {
                     Ok((amt, _src)) => {
                         let buf = &buf[..amt];
                         let _ = itx.send(Vec::from(buf));
-                        socket_last_success = ::time::SteadyTime::now();
+                        timeout = ::time::SteadyTime::now() + ::time::Duration::seconds(SOCKET_FAILURE_ALLOWANCE);
                     },
                     Err(ref e) if e.kind() == ErrorKind::TimedOut => {
                         // This extra loop ensures all pending messages are sent
@@ -50,7 +50,7 @@ impl UtpWrapper {
                                 },
                                 Err(TryRecvError::Disconnected) => break 'outer,
                                 Err(TryRecvError::Empty) => {
-                                    if socket_last_success + ::time::Duration::seconds(SOCKET_FAILURE_ALLOWANCE) < ::time::SteadyTime::now() {
+                                    if timeout < ::time::SteadyTime::now() {
                                         break 'outer;
                                     } else {
                                         break;
--- a/src/utp_wrapper.rs
+++ b/src/utp_wrapper.rs
@@ -50,18 +50,18 @@ impl UtpWrapper {
                                 Err(TryRecvError::Disconnected) => break 'outer,
                                 Err(TryRecvError::Empty) => {
                                     attempts += 1;
-                                    if attempts < 5000 {
+                                    if attempts < 10000 {
                                         break;
                                     } else {
                                         break 'outer;
                                     }
                                 }
                             }
+                            ::std::thread::sleep(::std::time::Duration::from_millis(1));
                         }
                     },
                     Err(_) => break,
                 }
-                ::std::thread::sleep(::std::time::Duration::from_millis(1));
             }
         }).unwrap();
         Ok(UtpWrapper {
--- a/src/utp_wrapper.rs
+++ b/src/utp_wrapper.rs
@@ -8,6 +8,7 @@ use std::net::SocketAddr;
 
 const CHECK_FOR_NEW_WRITES_INTERVAL_MS: i64 = 50;
 const BUFFER_SIZE: usize = 1000;
+const SOCKET_FAILURE_ALLOWANCE : i64 = 15;
 
 pub struct UtpWrapper {
     input: Receiver<Vec<u8>>,
@@ -26,7 +27,7 @@ impl UtpWrapper {
         socket.set_read_timeout(Some(CHECK_FOR_NEW_WRITES_INTERVAL_MS));
 
         let _ = thread::Builder::new().name("rust-utp multiplexer".to_string()).spawn(move || {
-            let mut attempts = 0;        
+            let mut socket_last_success = ::time::SteadyTime::now();        
             'outer:
             loop {
                 let mut buf = [0; BUFFER_SIZE];
@@ -35,7 +36,7 @@ impl UtpWrapper {
                     Ok((amt, _src)) => {
                         let buf = &buf[..amt];
                         let _ = itx.send(Vec::from(buf));
-                        attempts = 0;
+                        socket_last_success = ::time::SteadyTime::now();
                     },
                     Err(ref e) if e.kind() == ErrorKind::TimedOut => {
                         // This extra loop ensures all pending messages are sent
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @vinipsmaker

src/utp_wrapper.rs
```diff
@@ -47,12 +48,20 @@ impl UtpWrapper {
...
+                                    } else {
+                                        break 'outer;
+                                    }
+                                }
                             }
                         }
                     },
                     Err(_) => break,
                 }
+                ::std::thread::sleep(::std::time::Duration::from_millis(1));
```
vinipsmaker: Is this necessary? The `UdpSocket::read_from` blocking call will already put the thread to sleep and avoid unnecessary CPU share. Also, by sleeping the thread like this, you increase the time we're not reading from socket and therefore increase the chance we lose some packet.

src/utp_wrapper.rs
```diff
@@ -47,12 +48,20 @@ impl UtpWrapper {
                                     }
                                 },
                                 Err(TryRecvError::Disconnected) => break 'outer,
-                                Err(TryRecvError::Empty) => break,
+                                Err(TryRecvError::Empty) => {
+                                    attempts += 1;
+                                    if attempts < 5000 {
+                                        break;
+                                    } else {
+                                        break 'outer;
```
vinipsmaker: this change doesn't look cool, as it'll limit us to always put something in the "write queue" if we don't drop the connection. doesn't look good.

maqi: no you don't have to. the loop is around try_recv , which means if there is nothing in the queue, it will `break` eventually.
However, it maybe better change to use a timeout, but that shall be after CI confirms the change works.

src/utp_wrapper.rs
maqi: My local test shows it is necessary.
Let's see what CI builds say.

src/utp_wrapper.rs
vinipsmaker: It's okay to the `orx` to be `Empty`. It happens when the crust user hasn't scheduled any message to be written. I'm not understanding why this should trigger a connection loss.

vinipsmaker:
AppVeyor is failing.

vinipsmaker:
We shouldn't merge a test that can fail on `master`, as it'll make all new PRs subject to failure even if the code is correct. That's why we ignore some tests. Please add a `#[ignore]` directive to the new test.

src/utp_wrapper.rs
```diff
@@ -47,7 +49,13 @@ impl UtpWrapper {
                                     }
                                 },
                                 Err(TryRecvError::Disconnected) => break 'outer,
-                                Err(TryRecvError::Empty) => break,
+                                Err(TryRecvError::Empty) => {
+                                    if timeout < ::time::SteadyTime::now() {
```
vinipsmaker: could you elaborate on the intention of this change?

maqi: the outer loop (the socket connection thread) needs to be broken (existed) once there is continuously timed-out events for a long period.
this is the main point causing the hanging issue on windows as the dangling thread will keeps waiting forever.

maqi:
>We shouldn't merge a test that can fail on master, as it'll make all new PRs subject to failure even if the code is correct. That's why we ignore some tests. Please add a #[ignore] directive to the new test.

I prefer not to use `#[ignore] directive` as this is the main test to verify the rust-utp usage.
And, I hope with the work of PR https://github.com/maidsafe/crust/pull/445 and rust-utp updates, this failing test can be fixed soon.

src/utp_wrapper.rs
vinipsmaker: The timeout **always** happen, because we've set a timeout [here](https://github.com/maidsafe/crust/pull/442/files#diff-68671ba860f232f621493a8fac59d30fR38). This timeout purpose is to allow crust to break out of read cycle and check if there are pending messages to be sent.

What your code is doing is "drop the connection it the crust user don't schedule any messages to deliver within this amount of time".

maqi: Sorry, the link you put seems won't break the outer loop, i.e. stop the thread?
neither from the testing code I see any timeout happens for the outer loop.


The purpose of this PR is to resolve the hanging issue of the test, so the test itself can be included in the CI and to be used to verify the utp related work.
Yes, maybe as you said the change means stop the connection if there is no scheduled msg.
But I think that is design issue to decide whether it is valid or not.
Maybe a keepalive msg is necessary, and I see your work in https://github.com/maidsafe/crust/pull/445 introducing ping-pong msgs.

If you are concerned about the master branch, I will suggest merge this PR into your 445 work, then close this PR.
So PR 445 will contain all the testing code and updates required for the updated rust-utp.

canndrew:
Is there a reason the "rust-utp multiplexer" thread isn't two threads (one for input and one for output)? At the moment you've got

```basic
10 READ FOR 50ms
20 WRITE ANY DATA WE HAVE
30 GOTO 10
```

Which is pretty ugly. It's ugly that there's a mandatory ~50ms latency on all writes and it's ugly that the thread will be continuously waking up when it has nothing to do.

Also the connection will drop if it doesn't get used for 15 seconds..?

maqi:
@canndrew : i think it's better to be answered by someone else why it is one thread, maybe @vinipsmaker ?

The 15 seconds timeout is just I added in to resolve the hanging issue in CI builds on Windows.

canndrew:
> The 15 seconds timeout is just I added in to resolve the hanging issue in CI builds on Windows.

Are we sending some kind of keepalive packet on these connections? An `exit(0)` on the first line of code would also resolve the hanging issue.

maqi:
I am not sure the exist(0) will work, as the loop seems does need to keep alive when there is msgs flying.

I think @vinipsmaker has PR https://github.com/maidsafe/crust/pull/445 which introducing keep-alive style ping-pong msg. However I am not sure whether this is affect here.

canndrew:
> I am not sure the exist(0) will work, as the loop seems does need to keep alive when there is msgs flying.

Right, but does it also need to keep alive for more than 15 seconds when there aren't messages flying?

dirvine:
If it helps in routing when we connect to nodes we wish to stay connected until one side drops or goes off-line. So the connections remain until we kill them (long lived connections).

canndrew:
@dirvine That's kinda what I figured.

src/utp_wrapper.rs
vinipsmaker: > But I think that is design issue to decide whether it is valid or not.

I hardly think so. I can write a program to download a 10GB file from an HTTP server that will only, and only, read bytes after the initial HTTP boilerplate is written. This behaviour isn't wrong and shouldn't be forbidden.

> The purpose of this PR is to resolve the hanging issue of the test, so the test itself can be included in the CI and to be used to verify the utp related work.

Isn't there another way to achieve it?

vinipsmaker:
> Is there a reason the "rust-utp multiplexer" thread isn't two threads (one for input and one for output)?

https://github.com/meqif/rust-utp/issues/4

> At the moment you've got [...] Which is pretty ugly.

Agreed, but the issue should be fixed at https://github.com/meqif/rust-utp/issues/4 .

> It's ugly that there's a mandatory ~50ms latency on all writes and it's ugly that the thread will be continuously waking up when it has nothing to do.

No, it's not for **all** writes. Once we see a message on the write queue, we keep checking for all of them and sending them before we go back to the timeout cycle.

> Also the connection will drop if it doesn't get used for 15 seconds..?

I'm also struggling to understand this new code.

vinipsmaker:
> I am not sure the exist(0) will work, as the loop seems does need to keep alive when there is msgs flying.

If it helps, the channel is closed when the other end is dropped. So we can drop the sending/receiving channels from our threads within state and the thread can stop as soon as it receives an `Err` from the `recv` call.

src/utp_wrapper.rs
maqi: >> But I think that is design issue to decide whether it is valid or not.

>I hardly think so. I can write a program to download a 10GB file from an HTTP server that will only, and only, read bytes after the initial HTTP boilerplate is written. This behaviour isn't wrong and shouldn't be forbidden.

That's exactly a `design` issue, as there is such usage case needs to fulfilled, so our product `needs to be designed` so.

>> The purpose of this PR is to resolve the hanging issue of the test, so the test itself can be included in the CI and to be used to verify the utp related work.

>Isn't there another way to achieve it?

It will then require some changes to the utp_wrapper API and better to be `re-designed`. I think it shall be better to be left over to a better crust understanding guy, such as you and @vinipsmaker , and better for another PR.

madadam:
I'm not big fan of the timeouts introduced in this PR. They don't fix anything, they only work around the underlying cause. Also, I managed to replicate the hangup issue on linux too, it just happens much less frequently. So this isn't windows-only issue. @inetic seems to be encountering related issues in rust-utp, so it could be the actual cause of the hangups is there, not in crust.

src/utp_wrapper.rs
vinipsmaker: > It will then require some changes to the utp_wrapper API and better to be `re-designed`. I think it shall be better to be left over to a better crust understanding guy, such as you and @vinipsmaker , and better for another PR.

Okay, I've opened a [new PR to investigate the hanging issue under Windows](https://github.com/maidsafe/crust/pull/451).

PR #444, opened 2015-12-09
Fix build on Rust nightly

vinipsmaker:
Unfortunately, Rust decided to deprecate the feature we're using, so code is uglier now. But, at least, we're closer to build on stable Rust.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/444)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ log = "~0.3.3"
 crossbeam = "~0.1.6"
 memmap = "0.2.2"
 maidsafe_utilities = "~0.1.3"
+ip = "1.0.0"
 
 [dev-dependencies]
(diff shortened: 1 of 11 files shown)


maidsafe-highfive:
r? @inetic

Cargo.toml
```diff
@@ -25,6 +25,7 @@ log = "~0.3.3"
 crossbeam = "~0.1.6"
 memmap = "0.2.2"
 maidsafe_utilities = "~0.1.3"
+ip = "1.0.0"
```
Fraser999: Shouldn't we use `ip = "~1.0.0"`?  (same for memmap).  I guess that can be updated in the next PR to avoid blocking this one from being merged.

Issue #439, opened 2015-12-02
Failing of connection establishment and missing messages on rust-utp socket test


maqi:
As detailed described in the Rust-UTP PR https://github.com/maidsafe/rust-utp/pull/8

the test proves a failing rate at around : 0.005% when socket without timeout and 0.25% when socket with timeout.

according to the work of Crust PR https://github.com/maidsafe/crust/pull/437 , the test proves the crust level failing rate is around the same as the rust-utp when socket with timeout.

vinipsmaker:
I'm suspecting this is being caused by the congestion control algorithm (and the changes brought by the `timeout` branch). I'm going to investigate it in detail.

vinipsmaker:
I based my work on top of @maqi's `test_network_utp` [branch](https://github.com/maqi/crust/tree/b6f4af3819740b7c1d39ea30d3a5ab69f7133cfe).

After some attempts, debugging hours, observations and so on, I'm suspecting the problem is on the logic of the network. I believe the problem is on the congestion control algorithm from rust-utp. Missing packets make uTP unreliable. uTP isn't working properly under unreliable connections (which is exactly where UDP is supposed to be used). On the tests done, it was observed that the use of custom user timeout -- which is a feature used by crust to make concurrent reads/writes possible -- would increase the failure/hang rate.

My newest try was to change [`CHECK_FOR_NEW_WRITES_INTERVAL_MS`](https://github.com/maqi/crust/blob/b6f4af3819740b7c1d39ea30d3a5ab69f7133cfe/src/utp_wrapper.rs#L9) to support my claim/suspicion. Making the check more "granular" would increase the chance to not be reading when the packet was sent, which would cause a "lost packet". The "lost packet" is exactly the reason why I believe the test is failing.

Under the usual `50i64`, the rates I got from my machine were:

 ok | err
--- | ----
  6 |  15

When I changed the value to `200i64`, the stats were:

 ok | err
--- | ----
 11 |  10

With my suspicion confirmed (actually, I'd need to run more tests, but they take too long to complete and I already have something to change and test to see if the problem goes away), the next step is to study the congestion control algorithm of rust-utp/uTP and propose a solution to make it more reliable (and maybe give better guarantees on connection timed out). Until a definitive solution is implemented, crust could change the `CHECK_FOR_NEW_WRITES_INTERVAL_MS` value to decrease the rate failure and use ping/pong messages to detect connection loss itself.

[...]

PR #445, opened 2015-12-10
Improve uTP

vinipsmaker:
This updates #439 by adding more protection on the crust side of things. To close the issue for good, updates are also required on the rust-utp side.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/445)

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -18,10 +18,11 @@
 use std::collections::{HashMap, HashSet};
 use std::io;
 use std::sync::mpsc;
-use std::sync::mpsc::{Sender, Receiver};
+use std::sync::mpsc::{Sender, Receiver, TryRecvError};
 use std::thread;
 use std::boxed::FnBox;
 use std::str::FromStr;
+use std::time::Duration;
 
 use beacon;
 use bootstrap_handler::BootstrapHandler;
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @inetic

maqi:
could you link the required rust-utp updates as well ?
also, a detailed test report of the failure rate before and after this update will be preferred to be included as well.
otherwise it will be hard to tell whether this update is necessary.

Also, the PR https://github.com/maidsafe/crust/pull/442 is preferred to be merged before this PR,
as the testing code of network using utp is in that one.

vinipsmaker:
> could you link the required rust-utp updates as well ?

I started working on this after I submitted this PR. The idea is to not leave crust in a so bad situation while we fix the real issue within rust-utp.

> also, a detailed test report of the failure rate before and after this update will be preferred to be included as well.

Well, the current issues won't happen because if rust-utp misbehave, crust will issue a "connection loss" event. We "just" exchange one error for another. Connection loss is expected in the wild and layers above crust will have to handle them anyway. With the patch, they'll have the chance to not hang thanks to misbehaving rust-utp.

maqi:
>> could you link the required rust-utp updates as well ?

>I started working on this after I submitted this PR. The idea is to not leave crust in a so bad situation while we fix the real issue within rust-utp.

If so, I will prefer merge your rust-utp work first then merge this PR, or merge them at the same time.

>> also, a detailed test report of the failure rate before and after this update will be preferred to be included as well.

>Well, the current issues won't happen because if rust-utp misbehave, crust will issue a "connection loss" event. We "just" exchange one error for another. Connection loss is expected in the wild and layers above crust will have to handle them anyway. With the patch, they'll have the chance to not hang thanks to misbehaving rust-utp.

However, the current crust network test only covers TCP, and this`promised behaviour` is not tested. It is good crust raise a connection loss instead of just hanging, but we need a report here of such failing rate is acceptable.

PR #449, opened 2015-12-16
Fix simple_sender and simple_receiver examples.

canndrew:
Disabling the beacon from starting automatically caused these examples to stop working. Beacon is now started manually in both examples.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/449)


maidsafe-highfive:
r? @vinipsmaker

PR #451, opened 2015-12-28
Vinipsmaker test utp

vinipsmaker:
CIs will fail to build until we release the new version of rust-utp.

However, tests are running fine on my local machine.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/451)

PR #407, opened 2015-11-05
*WIP* Adds uTP network test (currently failing)

inetic:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/407)

PR #454, opened 2015-12-28
Adds uTP network test

vinipsmaker:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/454)

### DIFF
--- a/src/acceptor.rs
+++ b/src/acceptor.rs
@@ -94,6 +94,60 @@ impl Acceptor {
         })
     }
 
+    #[cfg(test)]
+    pub fn with_utp(listener: ::utp::UtpListener,
+                    hole_punch_server: Arc<HolePunchServer>,
+                    connection_map: Arc<ConnectionMap>)
+        -> io::Result<Acceptor> {
+        let running = Arc::new(AtomicBool::new(true));
+        let running_cloned = running.clone();
+        let addr = try!(listener.local_addr());
+        let mapped_addrs = try!(get_if_addrs())
+                                .into_iter()
+                                .filter(|i| !i.is_loopback())
+                                .map(|i| SocketAddr::new(i.ip(), addr.port()))
+                                .collect();
+        let joiner = RaiiThreadJoiner::new(thread!("acceptor-utp", move || {
+            loop {
+                let mapper_external_addr = hole_punch_server.external_address();
+                let mapper_internal_port = hole_punch_server.listening_addr().port();
+                let handshake = Handshake {
+                    mapper_port: Some(mapper_internal_port),
+                    external_addr: mapper_external_addr,
+                    remote_addr: SocketAddr(net::SocketAddr::from_str("0.0.0.0:0").unwrap()),
+                };
+                let accept_res = transport::accept_utp(&listener).and_then(|t| {
+                    transport::exchange_handshakes(handshake, t)
+                });
+                if !running_cloned.load(Ordering::SeqCst) {
+                    break;
+                }
+                // TODO (canndrew): What to do with this error?
+                let _ = match accept_res {
+                    Ok((handshake, transport)) => {
+                        let c = transport.connection_id.clone();
+                        let protocol = *c.peer_endpoint()
+                                         .protocol();
+                        let remote_addr = SocketAddr(*handshake.remote_addr);
+                        let our_external_endpoint = Endpoint::from_socket_addr(protocol, remote_addr);
+                        connection_map.register_connection(handshake, transport, Event::OnBootstrapAccept(our_external_endpoint, c))
+                    }
+                    Err(e) => {
+                        warn!("Acceptor got an error: {} {:?}", e, e);
+                        break;
+                    }
+                };
+            }
+        }));
+
+        Ok(Acceptor {
+            _joiner: joiner,
+            running: running,
+            addr: addr,
+            mapped_addrs: mapped_addrs,
+        })
+    }
+
     pub fn local_address(&self) -> SocketAddr {
         SocketAddr(self.addr)
     }
@@ -104,8 +158,16 @@ impl Acceptor {
 }
 
 impl Drop for Acceptor {
+    #[cfg(not(test))]
     fn drop(&mut self) {
         self.running.store(false, Ordering::SeqCst);
         let _ = TcpStream::connect(util::unspecified_to_loopback(&self.addr));
+    }
+
+    #[cfg(test)]
+    fn drop(&mut self) {
+        self.running.store(false, Ordering::SeqCst);
+        let _ = TcpStream::connect(util::unspecified_to_loopback(&self.addr));
+        let _ = ::utp::UtpSocket::connect(util::unspecified_to_loopback(&self.addr));
     }
 }
--- a/src/service.rs
+++ b/src/service.rs
@@ -1247,6 +1247,10 @@ mod test {
                                         stats.connect_count += 1;
                                         self.send_data_to(connection);
                                     }
+                                    Event::OnBootstrapConnect(Err(e), _) => {
+                                        thread::sleep(::std::time::Duration::from_secs(120));
+                                        panic!("Cannot establish all connections {:?}", e)
+                                    }
                                     Event::OnBootstrapAccept(_, connection) => {
                                         stats.accept_count += 1;
                                         self.send_data_to(connection);
--- a/src/service.rs
+++ b/src/service.rs
@@ -1235,11 +1235,13 @@ mod test {
 
             fn run(&mut self) -> Stats {
                 let mut stats = Stats::new();
-
-                for it in self.category_rx.iter() {
-                    match it {
-                        ::maidsafe_utilities::event_sender::MaidSafeEventCategory::CrustEvent => {
+                let timeout = ::time::SteadyTime::now() + ::time::Duration::milliseconds(120000);
+                loop {
+                    match self.category_rx.try_recv() {
+                        Ok(::maidsafe_utilities::event_sender::MaidSafeEventCategory::CrustEvent) => {
                             if let Ok(event) = self.reader.try_recv() {
+                                debug!("node {:?} received event : {:?} current msg counting {:?} among {:?}",
+                                       self._id, event, stats.messages_count, TOTAL_MSG_TO_RECEIVE);
                                 match event {
                                     Event::OnBootstrapConnect(Ok((_, connection)), _) => {
                                         stats.connect_count += 1;
@@ -1251,19 +1253,28 @@ mod test {
                                     }
                                     Event::NewMessage(_from, _bytes) => {
                                         stats.messages_count += 1;
-                                        // let msg = decode::<String>(&bytes);
                                         if stats.messages_count == TOTAL_MSG_TO_RECEIVE {
                                             break;
                                         }
                                     }
-                                    Event::LostConnection(_) => {}
+                                    Event::LostConnection(_) => {
+                                        println!("connection lost");
+                                        break;
+                                    }
                                     _ => {
-                                        println!("Received event {:?}", event);
+                                        println!("Received unknown event {:?}", event);
                                     }
                                 }
                             }
                         }
-                        _ => unreachable!("This category should not have been fired - {:?}", it),
+                        Ok(_) => unreachable!("This category should not have been fired"),
+                        Err(_) => {}
+                    }
+                    if ::time::SteadyTime::now() < timeout {
+                        let duration = ::std::time::Duration::from_millis(1);
+                        ::std::thread::sleep(duration);
+                    } else {
+                        break;
                     }
                 }
                 stats
--- a/src/service.rs
+++ b/src/service.rs
@@ -1258,8 +1258,7 @@ mod test {
                                         }
                                     }
                                     Event::LostConnection(_) => {
-                                        println!("connection lost");
-                                        break;
+                                        debug!("connection lost");
                                     }
                                     _ => {
                                         println!("Received unknown event {:?}", event);
--- a/src/service.rs
+++ b/src/service.rs
@@ -149,6 +149,22 @@ impl Service {
         Ok(Endpoint::from_socket_addr(Protocol::Tcp, SocketAddr(accept_addr)))
     }
 
+    #[cfg(test)]
+    pub fn start_accepting_utp(&mut self, port: u16) -> io::Result<Endpoint> {
+        use utp::UtpListener;
+        let listener = try!(UtpListener::bind(("0.0.0.0", port)));
+        let accept_addr = try!(listener.local_addr());
+
+        let hole_punch_server = self.mapper.clone();
+        let acceptor = try!(Acceptor::with_utp(listener, hole_punch_server,
+                                               self.connection_map.clone()));
+        self.acceptors.push(acceptor);
+
+        // FIXME: Instead of hardcoded wrapping in loopback V4, the
+        // listener should tell us the address it is accepting on.
+        Ok(Endpoint::from_socket_addr(Protocol::Utp, SocketAddr(accept_addr)))
+    }
+
     fn populate_bootstrap_contacts(&mut self,
                                    config: &Config,
                                    beacon_port: Option<u16>,
@@ -1185,8 +1201,7 @@ mod test {
         }));
     }
 
-    #[test]
-    fn network() {
+    fn test_network(protocol: Protocol) {
         BootstrapHandler::cleanup().unwrap();
 
         const NETWORK_SIZE: u32 = 10;
@@ -1269,7 +1284,12 @@ mod test {
         let mut runners = Vec::new();
 
         let mut listening_eps = nodes.iter_mut()
-                                     .map(|node| node.service.start_accepting(0).unwrap())
+                                     .map(|node| {
+                                         match protocol {
+                                             Protocol::Tcp => node.service.start_accepting(0).unwrap(),
+                                             Protocol::Utp => node.service.start_accepting_utp(0).unwrap(),
+                                         }
+                                     })
                                      .map(|ep| ep.unspecified_to_loopback())
                                      .collect::<::std::collections::VecDeque<_>>();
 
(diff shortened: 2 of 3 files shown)


maidsafe-highfive:
r? @inetic

vinipsmaker:
Continued work from #451

vinipsmaker:
This test seems to be hanging on Windows, but actually [the Windows support on `master` is already broken](https://ci.appveyor.com/project/MaidSafe-QA/crust/branch/master/job/wa6qhyoy3q99j2g7) and shows the exact same issue. In summary, this PR might not be affecting Windows support at all.

I'll leave it to you whether to accept/reject this PR as is or to fix the the Windows hanging issue prior to the addition of the `test-network-utp`.

vinipsmaker:
> I'll leave it to you whether to accept/reject this PR as is or to fix the the Windows hanging issue prior to the addition of the `test-network-utp`.

I changed my mind. [Test is failing on Windows](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.495/job/jp6ma92jgq1n58w0#L393). Better to fix one problem at a time instead accumulating'em all.

vinipsmaker:
[There is some issue under Windows](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.500/job/vh3l0p2ki0sg943f#L378):

    Received unknown event OnConnect(Err(Error { repr: Custom(Custom { kind: Other, error: StringError("handshake failed") }) }), 0)

vinipsmaker:
The [tracked down Windows error](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.500/job/vh3l0p2ki0sg943f#L465):

    Error setting read timeout: Error { repr: Os { code: 10093, message: "Either the application has not called WSAStartup, or WSAStartup failed." } }

Probable origin: https://github.com/maidsafe/rust-utp/blob/ffa531c0c58e078baa60c36661fd185825739813/src/socket.rs#L576

vinipsmaker:
> Probable origin: https://github.com/maidsafe/rust-utp/blob/ffa531c0c58e078baa60c36661fd185825739813/src/socket.rs#L576

I was wrong about that.

> `Error setting read timeout: Error { repr: Os { code: 10093, message: "Either the application has not called WSAStartup, or WSAStartup failed." } }`

I'm confused. It's like I'm getting a different error every time.

- [`"A blocking operation was interrupted by a call to WSACancelBlockingCall."`](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.502/job/3ri31eiwwm66to4b#L447).
- [`"A message sent on a datagram socket was larger than the internal message buffer or some other network limit, or the buffer used to receive a datagram into was smaller than the datagram itself."`](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.503/job/jejrsxa1k5twjli4#L400).

Windows support seems to be much less robust than Linux/OSX.

vinipsmaker:
I investigated the issue more and [tracked down to this function](https://github.com/maidsafe/rust-utp/blob/ffa531c0c58e078baa60c36661fd185825739813/src/socket.rs#L595), which is weird. It's weird because [the error is `"An existing connection was forcibly closed by the remote host."`](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.504/job/ks2vv7qfg9v7sgyw#L390).

UDP is connection-less, so why this error? I found [an explanation on StackOverflow](http://stackoverflow.com/questions/30749423/is-winsock-error-10054-wsaeconnreset-normal-with-udp-to-from-localhost#comment49588739_30749423). So I think the correct behaviour is to ignore this errors, as UDP is connection-less and not being able to receive now doesn't mean not able to receive a few seconds later. I'll make a test and find out.

vinipsmaker:
> I'll make a test and find out.

Just did a small test and it was not enough, but I didn't get much info. Now I'll investigate `Err(e)` values from the `send_to` function.

vinipsmaker:
A part of the crust workflow in the introduced test is like (note that some parts only exist in crust and the test will indirectly and implicitly invoke such actions):

1. Establish a uTP connection (involves exchanging UDP messages). This is happening successfully.
2. Moves the uTP connection to another thread that will continuously wait for new UDP/uTP messages. This is failing and eventually the connection times out.
3. Start handshake exchange to confirm the new connection establishment. Then expose the connection to the user (in this case the test).
4. The connection times out, so no connection is confirmed and the test fails without even exchanging all messages.

I can confirm the uTP connection is established as the multiplexer thread is created. I only got this far by making rust-utp ignore the error codes 10054 (WSAECONNRESET) and 10040 (WSAEMSGSIZE) on Windows.

WSAECONNRESET looks safe to ignore, but WSAEMSGSIZE is really weird.

For now, I'll research if moving UDP socket descriptors among threads can cause any problem on Windows.

And it's troublesome to debug some error without having direct access to the operating system. Maybe next year I'll get a new laptop and install Windows on my current/old laptop.

vinipsmaker:
> For now, I'll research if moving UDP socket descriptors among threads can cause any problem on Windows.

Before continuing to investigate this, [I did one more test just to confirm that the errors only start after the objects are moved among threads](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.518/job/c4dka2tmm5vb2reg#L479). Now my suspicion has grow stronger.

vinipsmaker:
- I haven't found much (actually I found nothing) on StackOverflow (or Google...) about threading and UDP problems under Windows.
- I stopped getting 10040 error after increasing buffer size.
- There are some `"Failed to decode CBOR"` errors happening before connection times out. This is one thing I could investigate.
- I could also refactor crust support to rust-utp to restrict the "uTP thread" to be the same from connection establishment until application exit. And then check if the error persists.

vinipsmaker:
After a **long** time debugging, [I found something interesting](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.526/job/t6g2c85xgsso8pn2#L395).

`receive_handshake` from `transport.rs` fails at CBOR decoding because `"Missing field \'remote_ip\' in map object."`. Now I need to know if it's a bug thanks to corrupted stream from rust-utp or if it is the act of sending of a wrong structure within crust side (which would be very obscure, as TCP is working and the error only happens on Windows).

Also, even if we're facing this bug now, it doesn't mean it is THE bug, as other changes are required to make crust working properly under Windows for UDP support (ignoring `10054` error...).

I think it's unlikely I'll finish the fix before Jan 4th, but maybe next week I can bury these bugs for good.

vinipsmaker:
I found something interesting.

First a summary of the important part of the test to understand my finding:

1. The test create several nodes.
2. Each node will:
  1. Repeat the following for every other node:
    1. Connect to other node.
      1. Exchange handshake.
    2. Send message (it's coded as a `UserBlob` variant somewhere deeper at crust level).
    3. Wait for message to arrive.
    4. Once all messages are received, exit.

Now consider only a pair of nodes. We can have the following:

1. First node is "faster" (race?). It'll send handshake, receive handshake and then send `UserBlob`.
2. [The second node will try to receive a handshake (which should be the first CBOR structure that is passed in the channel), but it receives a `UserBlob`](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.531/job/nhblrgwiax096u6y#L793).
  1. [The `remote_ip field missing`](https://github.com/BurntSushi/rust-cbor/blob/c4336891b37eab1b0e22edb2d8de8e946a7ccd5f/src/rustc_decoder.rs#L297).
  2. I actually changed `rust-cbor` to print the value of the `map` variable when this error happens. In the CI output, this is `{"fields": Array([Array([Unsigned(UInt8(105)), Unsigned(UInt8(77)), Unsigned(UInt8(69)), Unsigned(UInt8(83)), Unsigned(UInt8(83)), Unsigned(UInt8(65)), Unsigned(UInt8(71)), Unsigned(UInt8(69)), Unsigned(UInt8(32)), Unsigned(UInt8(48))])]), "variant": Unicode("UserBlob")}`.

Some observations:

- The error only happens on Windows.
- Several connections succeed and handshake is established successfully.
- The `rust-utp` stream is **NOT** corrupted. That's why CBOR is successfully decoded (but an unexpected structure is found). So it's very very very unlikely this error is happening on `rust-utp` side. At least initially, we should focus on debugging on the `crust` side.

In short, there might be a race within `crust`. `UserBlob` is being written/received before handshake.

vinipsmaker:
@inetic mentioned that he also was getting out-of-order packets (at least I understood this) and (implicitly) suggested me to try his fork: https://github.com/inetic/rust-utp/tree/master

But using his fork I got this error (WSAEMSGSIZE) when calling `recv_from` within `utp_wrapper.rs`: https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.532/job/pd4ehiqnngm643mh#L781

Then I proceed to merge my changes and these other two @inetic's commits:

- https://github.com/inetic/rust-utp/commit/1a578c6f70afc286faa1fe507a0cbdb08db2c76e
- https://github.com/inetic/rust-utp/commit/66e5b716a8c9d19ee54c56873bc451ff78088c73

And the test passed (but CI machine hanged). I'll do a few more tests before cleaning the code for release.

vinipsmaker:
[I got a new error, `EndOfStream`](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.534/job/7mxqgihsm2gi4277#L860), which is weird because the connection was supposed to be established at this point. Gotta try disable some of @inetic's commits to check if error persists.

vinipsmaker:
- Without commit https://github.com/inetic/rust-utp/commit/1a578c6f70afc286faa1fe507a0cbdb08db2c76e, out-of-order packets issue is **NOT** fixed: https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.535/job/tm75pqkvdp9dyxq8#L684
- Without commit https://github.com/inetic/rust-utp/commit/66e5b716a8c9d19ee54c56873bc451ff78088c73, out-of-order packets issue is **NOT** fixed either: https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.536/job/mer5ia16737j75m7#L728

So... both commits are required to fix the out-of-order packets issue, but then we have a new issue to fix.

vinipsmaker:
> So... both commits are required to fix the out-of-order packets issue, but then we have a new issue to fix.

Just a quick correction. Actually, I was seeing an `UserBlob` before a `Handshake`, but I have not verified that a `Handshake` was being received, so it could very well be a cause of a lost packet being incorrectly handled and corrupting stream (i.e. a bug within rust-utp). After discussing with @inetic, he argued that he did not fix an out-of-order issue. Anyways...

vinipsmaker:
> I stopped getting 10040 error after increasing buffer size.

I checked again and I saw that [actually still happens](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.549/job/5rjgyhwdr8gkbq0e#L903). It just happens to be very rare.

Also, it doesn't prevent the test from succeeding. Therefore, I reverted the commit that increases buffer size and tested again. The frequency of this error increased, but again, it didn't prevent the test from succeeding. The commit will remain disabled, as it represents no improvement.

vinipsmaker:
>> So... both commits are required to fix the out-of-order packets issue, but then we have a new issue to fix.

> Just a quick correction. Actually, I was seeing an `UserBlob` before a `Handshake`, but I have not verified that a `Handshake` was being received, so it could very well be a cause of a lost packet being incorrectly handled and corrupting stream (i.e. a bug within rust-utp). After discussing with @inetic, he argued that he did not fix an out-of-order issue. Anyways...

I've prepared and submitted a PR with the Windows-specific changes to the upstream project: https://github.com/meqif/rust-utp/pull/24

Together with @inetic's changes, `rust-utp` will be in a far better place.

With changes from both sources combined, failure rate went down to 1 out of 10. So I'm gonna focus on the Windows hanging problem for now (which is probably just a leaking thread problem).

vinipsmaker:
I found out which thread is leaking and causes the test under Windows to hang.

I've put some `println!` statements to let me know when the [reader thread](https://github.com/maidsafe/crust/blob/b4ec3f4109e8332d40dbd42b1e374c76a41df964/src/state.rs#L325) and the [writer thread](https://github.com/maidsafe/crust/blob/b4ec3f4109e8332d40dbd42b1e374c76a41df964/src/state.rs#L307) were about to exit.

Within the `test_network_tcp` test, all 180 threads exited. [Within the `test_network_utp` test, only the `writer` thread would exit. All `reader` threads remained running.](https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.569/job/4b9a2xdb5g57wspj#L634)

This is an issue that @madadam is already working on, as part of his https://github.com/maidsafe/crust/pull/466#issuecomment-169688321 PR. I'll wait for him to finish and work on something else and then come back later to test if his changes fix the test.

Once the hanging problem is fixed, there are a few more things to do with this test. For instance, I don't think commit https://github.com/vinipsmaker/crust/commit/e47bda86f82cf9867312ca1a10137050ee480f95 is necessary.

vinipsmaker:
- One of a set of patchsets necessary to make this PR pass merged: https://github.com/maidsafe/crust/pull/475
- https://github.com/maidsafe/rust-utp/pull/15 opened

vinipsmaker:
https://github.com/maidsafe/rust-utp/pull/15 merged.

Next one will be https://github.com/maidsafe/rust-utp/pull/14 and it should be enough to merge this PR with a test showing that uTP is stable and usable now.

vinipsmaker:
My progress in this task is stalled until [branch `master` is fixed](https://github.com/maidsafe/crust/issues/482).

vinipsmaker:
> My progress in this task is stalled until [branch `master` is fixed](https://github.com/maidsafe/crust/issues/482).

I fixed branch `master` and now I'm back to this problem.

vinipsmaker:
uTP seems stable when we use [Peter's patchset](https://github.com/maidsafe/rust-utp/pull/14).

However, I went a little further to be sure just in case:

- Executing `rust-utp`'s own tests.
  - On upstream's `master` branch: https://gist.github.com/vinipsmaker/753ff2b8b2b58f4250e6
  - On our fork's `master` branch (with some tests ignored because I was tired of waiting): https://gist.github.com/vinipsmaker/1f3315c021e7c97bf329
  - On Peter's new branch: https://gist.github.com/vinipsmaker/c905327cc1d5cfa79ecc

I'll investigate why the new changes make some of the old tests fail before merging the branch.

vinipsmaker:
Just because I'm used to find races, I've run the `rust-utp` tests again a few more times to be sure only the following tests were failing:

- `socket::test::test_acks_on_socket`
- `socket::test::test_handle_packet`
- `socket::test::test_receive_unexpected_reply_type_on_connect`

So... I'll start investigating one of these.

vinipsmaker:
My work on https://github.com/maidsafe/rust-utp/pull/16 PR is done. So, as soon as that PR is merged, this very one can also be merged.

vinipsmaker:
https://github.com/maidsafe/rust-utp/pull/16 was merged.

I added a dummy commit to trigger a new CI build, but it seems the uTP test failed on Windows:

- https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.688/job/tk8ketcuot9vcnfd#L404
- https://ci.appveyor.com/project/MaidSafe-QA/crust/build/1.0.688/job/cxxm36a68gqod2xn#L404

I'll have to investigate why.
:disappointed:

vinipsmaker:
> I'll have to investigate why.

It seems if the `bugfixes2` branch from `rust-utp` is used alone, tests pass. However, if `rust-utp` branch is merged into `master`, something goes terribly wrong. I'll start the investigation from here (which commit from `master` that is not on `bugfixes2` causes the failure?).

vinipsmaker:
>> I'll have to investigate why.

> It seems if the `bugfixes2` branch from `rust-utp` is used alone, tests pass. However, if `rust-utp` branch is merged into `master`, something goes terribly wrong. I'll start the investigation from here (which commit from `master` that is not on `bugfixes2` causes the failure?).

I was wrong. The CI builds on this PR failed because I forgot to release a new crate for `temp_utp`. `Cargo.toml` gets the latest released crate, not the latest commit on the `master` branch.

CI tests are passing now. PR is ready to be merged.

src/service.rs
```diff
@@ -149,6 +149,22 @@ impl Service {
...
+        use utp::UtpListener;
+        let listener = try!(UtpListener::bind(("0.0.0.0", port)));
+        let accept_addr = try!(listener.local_addr());
+
+        let hole_punch_server = self.mapper.clone();
+        let acceptor = try!(Acceptor::with_utp(listener, hole_punch_server,
+                                               self.connection_map.clone()));
+        self.acceptors.push(acceptor);
+
+        // FIXME: Instead of hardcoded wrapping in loopback V4, the
```
dirvine: Agreed the return of the bind should be used in these cases Ok for now though

src/transport/mod.rs
```diff
@@ -134,6 +134,28 @@ pub fn accept(acceptor: &TcpListener) -> IoResult<Transport> {
     })
 }
 
+// FIXME: There needs to be a way to break from this blocking command.
+// (Though this seems to be impossible with the current Rust TCP API).
+#[cfg(test)]
+pub fn accept_utp(acceptor: &::utp::UtpListener) -> IoResult<Transport> {
+    let (stream, _remote_endpoint) = try!(acceptor.accept()
+                                          .map_err(|e| io::Error::new(io::ErrorKind::NotConnected, e.description())));
```
dirvine: A lot of these changed in new crust code so soon we need proper error types that can be passed up. So the string type errors will become less used in the code.

src/acceptor.rs
```diff
@@ -104,8 +158,16 @@ impl Acceptor {
 }
 
 impl Drop for Acceptor {
+    #[cfg(not(test))]
+    fn drop(&mut self) {
+        self.running.store(false, Ordering::SeqCst);
```
dirvine: This is better handled using the net::Shutdown Paradigm. You can use this example https://github.com/maidsafe/crust/blob/new_refactor/src/connection.rs#L113

vinipsmaker:
@dirvine: all these pieces of code you're commenting related to make an uTP listener just mirror the **current** TCP listener in `master`. Code is almost the same for current TCP. And I think it's wise to leave as is to avoid confusion.

I'll create a new merge/PR for the `new_refactor` branch after this PR is merged.

dirvine:
Sounds good, we are still toiling with compilation in new_refactor branch but yes good idea. Merging now.

vinipsmaker:
This one took quite a while:

- https://github.com/maidsafe/crust/pull/407 (5 Nov 2015)
- https://github.com/maidsafe/crust/pull/442
- https://github.com/maidsafe/crust/pull/451

And at least three developers have worked on this issue.

PR #460, opened 2016-01-06
Crust now compiles on Rust stable

madadam:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/460)

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -20,7 +20,6 @@ use std::io;
 use std::sync::mpsc;
 use std::sync::mpsc::{Sender, Receiver};
 use std::thread;
-use std::boxed::FnBox;
 use std::str::FromStr;
 
 use beacon;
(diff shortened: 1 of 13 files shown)


maidsafe-highfive:
r? @inetic

vinipsmaker:
Sorry, @inetic, gonna review this myself, so you have more time to fix your rust-utp stabilization task, which is something I want delivered as soon as possible.

vinipsmaker:
Can you add a new commit changing [.travis.yml](https://github.com/maidsafe/crust/blob/master/.travis.yml) to comment the "nightly" line and uncomment the "stable" line?

vinipsmaker:
PS: and the same for [appveyor.yml](https://github.com/maidsafe/crust/blob/master/appveyor.yml) too?

madadam:
@vinipsmaker done.

src/state.rs
```diff
@@ -41,7 +40,29 @@ use hole_punching::HolePunchServer;
...
+// dirty tricks, because Box<FnMut> is fine to call.
+//
+// This workaround can be removed once FnBox becomes stable or Box<FnOnce>
+// becomes usable.
+pub struct Closure(Box<FnMut(&mut State) + Send>);
+
+impl Closure {
+    pub fn new<F : FnOnce(&mut State) + Send + 'static>(f: F) -> Closure {
+        let mut f = Some(f);
+        Closure(Box::new(move |state: &mut State| {
```
vinipsmaker: This seems cryptic for me. Care to elaborate?

vinipsmaker:
Rest of the PR seems fine and tests passed. Looking forward to merge it.

src/state.rs
madadam: Yeah. Basically we need a closure that implements `FnOnce` to be stored in a `Box`. This is not usable in the current stable rust, because you can't call it back. That's why `FnBox` was introduced, but it is behind a unstable feature gate, so we can't use it either. The workaround is basically to use `Box<FnMut>` instead, but to be able to store `FnOnce`'s in it, I need to put that `FnOnce` into `Option` and then take it out when calling it. There is a lengthy and detailed discussion about this issue here: https://github.com/rust-lang/rust/issues/28796

The `Closure` struct is there just to wrap all this hackery in a convenient interface.

vinipsmaker: Thanks.

vinipsmaker: Just finished reading the thread and went to read your code again. Clever solution.

Issue #431, opened 2015-11-20
Send return 


dirvine:
I think if there is a send request to an invalid connection crust should return the error, atm it silently ignores the send request. This will change the API but provide valid info to the client of crust.

ustulation:
This crate had a major re-write/re-design. Please raise an issue again if required as almost everything in the library and the way it handled things have changed. Closing this.

Issue #405, opened 2015-11-05
uTP connection not detected as having been closed


Fraser999:
While running crust_peer on Ubuntu 15.10, if two instances form a uTP connection and one peer is then stopped, the other peer doesn't report the connection as having been broken.  I left the remaining instance to run for over 20 minutes after killing the peer, and on running `list` the dead peer still shows up as connected.

vinipsmaker:
Actually I'm having this issue with `TCP` as well. I might very well be just an issue with `crust_peer`. I'll take a deeper look now.

vinipsmaker:
Sorry, I've been mislead by the fact that connections aren't automatically removed from the list (must call `clean` on the CLI).

vinipsmaker:
After changing `crust_peer` to listen on `uTP`, here are my findings:

- Connect has dropped (`Lost connection to peer at Connection(Utp V4(0.0.0.0:58660) -> V4(127.0.0.1:56194))`) much much sooner than your 20 minutes. I guess the error you initially mentioned was related to issues fixed in `temp_utp` 0.7.0.
- Connection loss is not detect as fast as TCP. I believe the main thread closes before the uTP thread has a chance to send the `fin` packet. I'll investigate this problem.

madadam:
I need help with reproducing this issue. I'm bit confused about how to use crust_peer. When I run two instances, they automatically connect using TCP. How to I connect them using uTP?

On the other hand, if this issue is caused by UtpWrapper's thread not being joined, then that should be easy to fix. But I need some way to verify that the fix works.

madadam:
Ok, figured out how to reproduce the issue. Joining the UtpWrapper thread does not fix it though :(. My guess is that when the instance is killed forcefully, the destructors are not run, so no `fin` packet is sent. Going to investigate this a bit more to figure out the best way to fix it.

madadam:
Ok, so after a bit of playing with this, these are my observations:

- The uTP connection is dropped eventually, it just takes much more time than in the TCP case
- Joining the UtpWrapper thread on drop does not have any effect on the above

So I'm not sure there actually is a bug? I can push the UtpWrapper thread joining, which is something we should be doing anyway (it's not nice to keep hanging threads around), but otherwise not sure what else can be done.

madadam:
Should be fixed by #466.

vinipsmaker:
Yes, it should be fixed. Can you try again, @Fraser999?

Fraser999:
[...]

PR #466, opened 2016-01-07
Fixes dangling threads

madadam:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/466)

### DIFF
--- a/examples/crust_peer.rs
+++ b/examples/crust_peer.rs
@@ -669,7 +669,6 @@ fn main() {
     }
 
     drop(service);
-
     assert!(handler.join().is_ok());
 }
 
--- a/src/service.rs
+++ b/src/service.rs
@@ -118,11 +118,11 @@ impl Service {
 
     fn start_broadcast_acceptor(&mut self, beacon_port: u16) -> io::Result<u16> {
         let acceptor = try!(beacon::BroadcastAcceptor::new(beacon_port));
-        let b_port = acceptor.beacon_port();
+        let beacon_port = acceptor.beacon_port();
 
         // Right now we expect this function to succeed only once.
         assert!(self.beacon_guid_and_port.is_none());
-        self.beacon_guid_and_port = Some((acceptor.beacon_guid(), b_port));
+        self.beacon_guid_and_port = Some((acceptor.beacon_guid(), beacon_port));
 
         let sender = self.cmd_sender.clone();
 
@@ -137,7 +137,7 @@ impl Service {
         // TODO: Handle gracefuly.
         assert!(thread_result.is_ok());
 
-        Ok(b_port)
+        Ok(beacon_port)
     }
 
     /// This method tries to connect (bootstrap to existing network) to the default or provided
@@ -170,14 +170,15 @@ impl Service {
                                     blacklist: &[Endpoint]) {
         let config = self.config.clone();
         let beacon_guid_and_port = self.beacon_guid_and_port.clone();
-        let blist = blacklist.to_vec();
+        let blacklist = blacklist.to_vec();
 
-        Self::post(&self.cmd_sender, move |state: &mut State| {
-            let mut contacts = state.populate_bootstrap_contacts(&config,
-                                                                 beacon_port,
-                                                                 &beacon_guid_and_port);
+        Self::post(&self.cmd_sender, move |state : &mut State| {
+            let mut contacts = state.populate_bootstrap_contacts(
+                                    &config,
+                                    beacon_port,
+                                    &beacon_guid_and_port);
 
-            contacts.retain(|endpoint| !blist.contains(&endpoint));
+            contacts.retain(|endpoint| !blacklist.contains(&endpoint));
 
             state.bootstrap_off_list(token, contacts.clone());
         });
@@ -213,14 +214,14 @@ impl Service {
             // Connect to our listening ports, this should unblock
             // the threads.
             for port in &state.listening_ports {
+                let addr = ::util::loopback_v4(*port).get_address();
+
                 match *port {
                     Port::Tcp(_) => {
-                        let _ = TcpStream::connect(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1),
-                                                                     port.number()));
+                        let _ = TcpStream::connect(SocketAddrV4::new(addr, port.number()));
                     }
                     Port::Utp(_) => {
-                        let _ = UtpSocket::connect(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1),
-                                                                     port.number()));
+                        let _ = UtpSocket::connect(SocketAddrV4::new(addr, port.number()));
                     }
                 }
             }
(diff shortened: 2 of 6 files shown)


maidsafe-highfive:
r? @vinipsmaker

vinipsmaker:
This PR was created to address this issue: https://github.com/maidsafe/crust/issues/405#issuecomment-169263055

vinipsmaker:
> Joining the UtpWrapper thread does not fix it though

You're joining the `UtpWrapper` thread, but there is more.

[When a connection is registered (the `UtpWrapper` object), the handle is added to the `State::connections` collection and two new threads are created](https://github.com/maidsafe/crust/blob/b4ec3f4109e8332d40dbd42b1e374c76a41df964/src/state.rs#L258). As all the communication is done using `mpsc::channel`, you just need to destroy one end of the channel to force the other end to also close (and the thread will eventually stop). You should update `Service::stop` to also take care of removing all connections.

madadam:
Ok, I'll take a look at it. But even if we make sure we join all the threads that are spawned inside the service (and any objects used by it), it still won't solve the problem in #405. Because there are still multiple ways the connection can be broken without the `fin` packet being sent. For example:

- the peer process is killed forcefully (`kill - 9` ...)
- the peer machine is turned off, disconnected from power socket, physically destroyed, etc...
- the network goes down (router failure, power failure, cable being severed somewhere, etc...)
- and so on.

In none of this cases has the peer being disconnected chance to send the `fin` packet.

Anyway, even if this PR does not fix the problem in #405, I thinks it's still useful because I believe joining threads we spawn is something we should always be doing anyway.

dirvine:
> I believe joining threads we spawn is something we should always be doing anyway.

To me we should use crossbeam::scope for this and similar things where we can. Just for future anyway

madadam:
> To me we should use crossbeam::scope for this and similar things where we can. Just for future anyway

It seems that `crossbeam::scope` can only be used locally, inside a function. In this case we would need the crossbeam scope to be a member of a struct. We had a discussion on slack with @ustulation , @canndrew and @afck and the consensus was that crossbeam can't do that.

dirvine:
Yws, if we want more than that there is a scoped threadpool (crates.io) as well which is not function local.  Carrying a joinguard across functions may appear an anti pattern in many ways.

vinipsmaker:
> But even if we make sure we join all the threads that are spawned inside the service (and any objects used by it), it still won't solve the problem in #405.

Well, the problem was that we weren't sending the `fin` packet. It's okay to not receive a `fin` packet, as the connection will just time out and the other peer will drop the connection anyway (it just takes a little longer).

madadam:
Ah, I see. So even if the service was terminated gracefully, it still wouldn't send the fin packet, correct? So I guess this PR then fixes it? If it does, let me just add an explicit test case for it and then you can merge it if you think it's ok.

vinipsmaker:
> So I guess this PR then fixes it?

Not yet, do the changes I requested. The uTP thread might finish earlier/later. It's a race. If application exist, more likely it won't.

madadam:
You mean joining the [reader / writer threads](https://github.com/maidsafe/crust/blob/b4ec3f4109e8332d40dbd42b1e374c76a41df964/src/state.rs#L295) associated with each connection in `State` when it is stopped/destroyed?

vinipsmaker:
> You mean joining [...]

This would stall the application, you must close the associated channels (by dropping the connections) to request the threads to also exit.

madadam:
Yeah, but the connections are dropped automatically when the `State` is dropped, aren't they? So it should be enough to make sure the threads are joined after the connections are dropped, shouldn't it?

vinipsmaker:
> Yeah, but the connections are dropped automatically when the `State` is dropped, aren't they? So it should be enough to make sure the threads are joined after the connections are dropped, shouldn't it?

I'm seeing a deadlock. `State` joins the read/write threads. Connections aren't dropped (will be dropped later). `join` never returns, because threads don't finish, because connections weren't dropped.

It's possible to join the threads at other places anyway, so this can be avoided.

madadam:
That's why I want to join the threads only *after* the connection are dropped. Yup, it shouldn't be too hard. Have to run now, but will push a fix tomorrow.

madadam:
Fixed the dangling reader threads. Was caused by a `UtpWrapper` keeping a copy of output `Sender` which prevented the corresponding `Receiver` in the `UtpWrapper`'s thread to be closed.

inetic:
Can we please wait with merging this one until we have the `test_network_utp` test in place? I have the test locally passing with [this PR](https://github.com/maidsafe/rust-utp/pull/13), but it also exposes the leaky thread problem. So I tried to apply this PR, but then the `test_network_utp` test started failing again.

vinipsmaker:
> So I tried to apply this PR, but then the `test_network_utp` test started failing again.

I can investigate this issue. But I believe the problem is not this PR per si, but the crust side of things. Asio has a callback in the write operation, so you can choose to only really stop the service once TCP acknowledges proper delivery. The nodes from the test are like "when you receive a connection, send a message back and when you receive N messages, don't bother waiting for the pending messages to be delivered". Related to https://github.com/maidsafe/crust/issues/431 and how currently the "on-connect" event is delivered (versus the past version of it). This is just my suspicion, I need to properly investigate it.

> Can we please wait with merging this one until we have the `test_network_utp` test in place?

Sure we can wait, I can cherry-pick these commits and test locally to see if the new error is caused by this PR or by some other place.

In general, I think this PR is an improvement and there is nothing wrong with it. But waiting won't cause any trouble.

vinipsmaker:
> This branch has conflicts that must be resolved

Please update your work, @madadam.

madadam:
Fixed the merge conflicts.

vinipsmaker:
Please undo [these style changes](https://github.com/madadam/crust/commit/49a621f0640943d2211a54aaa69800a451ebb154#diff-bfa17b8ec2b0aac77da5f96cbcbe4101R49) or @dirvine will make another commit that will force every PR to be rebased again.
😛

vinipsmaker:
If you need help rewriting the commits, take a look into `git rebase -i`. It's a life-saver in these cases.

madadam:
I'm just going to apply `rustfmt` to my changes.

vinipsmaker:
> This branch has conflicts that must be resolved

madadam:
Fixed the conflicts, but some tests are failing. There are also some `#[ignore]`'d tests which should be checked. I won't be able to maintain this PR anymore as I've been moved to another project. Good luck ;)

vinipsmaker:
Gonna skip the following commits, as I believe they are just noise:

- https://github.com/madadam/crust/commit/8e5fd62508270510b06ed83b7c6b7e06b4eceaae
- https://github.com/madadam/crust/commit/417420b15f549d01f7a149a6efb8bb96092f7078

Also, continued your work on another PR, just to not having any conflicts with `master` at all while adding as little noise as possible: https://github.com/maidsafe/crust/pull/475

PR #467, opened 2016-01-07
code cleanup

vinipsmaker:
using nicer API that I learned because Adam Cigánek[1]

[1] https://github.com/maidsafe/crust/pull/460

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/467)


maidsafe-highfive:
r? @inetic

vinipsmaker:
The rendezvous test fails sometimes. CI failure under one of the tests is not caused by changes from this PR.

PR #468, opened 2016-01-07
General tidy up, lots to do, many unwraps to be handled

dirvine:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/468)

### DIFF
--- a/src/hole_punching.rs
+++ b/src/hole_punching.rs
@@ -40,9 +40,10 @@ pub struct SetExternalAddr {
     pub addr: SocketAddrW,
 }
 
-pub fn blocking_get_mapped_udp_socket(request_id: u32, helper_nodes: Vec<SocketAddr>)
-        -> io::Result<(UdpSocket, Option<SocketAddr>, Vec<SocketAddr>)>
-{
+pub fn blocking_get_mapped_udp_socket
+                                      (request_id: u32,
+                                       helper_nodes: Vec<SocketAddr>)
+                                       -> io::Result<(UdpSocket, Option<SocketAddr>, Vec<SocketAddr>)> {
     const MAX_DATAGRAM_SIZE: usize = 256;
 
     let udp_socket = try!(UdpSocket::bind("0.0.0.0:0"));
@@ -56,9 +57,13 @@ pub fn blocking_get_mapped_udp_socket(request_id: u32, helper_nodes: Vec<SocketA
     };
 
     let res = try!(::crossbeam::scope(|scope| -> io::Result<Option<(SocketAddr, usize)>> {
-        for helper in helper_nodes.iter() {
+        for helper in &helper_nodes {
             let sender = try!(udp_socket.try_clone());
-            let _periodic_sender = PeriodicSender::start(sender, *helper, scope, &send_data[..], ::std::time::Duration::from_millis(300));
+            let _periodic_sender = PeriodicSender::start(sender,
+                                                         *helper,
+                                                         scope,
+                                                         &send_data[..],
+                                                         ::std::time::Duration::from_millis(300));
             let deadline = ::time::SteadyTime::now() + ::time::Duration::seconds(2);
             let res = try!((|| -> io::Result<Option<(SocketAddr, usize)>> {
                 loop {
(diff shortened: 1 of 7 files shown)


maidsafe-highfive:
r? @inetic

src/service.rs
```diff
@@ -255,12 +258,12 @@ impl Service {
                             let _ = cmd_sender.send(Closure::new(move |state: &mut State| {
                                 let _ = state.handle_connect(token, h, t);
                             }));
-                        },
+                        }
```
vinipsmaker: removing comma here seems suspicious

vinipsmaker:
Please rewrite your work on top of `master`, so CI machines won't reject code because unstable Rust

src/service.rs
```diff
@@ -255,12 +258,12 @@ impl Service {
...
                         Err(e) => {
                             let _ = cmd_sender.send(Closure::new(move |state: &mut State| {
                                 let _ = state.event_sender.send(Event::OnConnect(Err(e), token));
                             }));
-                        },
+                        }
```
vinipsmaker: removing comma here doesn't seem suspicious, but the comma is a nice marker for the end of case and works better at merging patches (because if patch add a new case at the end, it doesn't need to modify previous line and I believe handing multiple of these patches is easier to automate if they don't try to change previous line so I believe git has a better time if we keep the ending comma)

src/service.rs
```diff
@@ -303,18 +307,18 @@ impl Service {
...
             let _ = Self::new_thread("rendezvous connect", move || {
-                match State::rendezvous_connect(handshake.clone(), udp_socket,
-                                                public_endpoint) {
+                match State::rendezvous_connect(handshake.clone(), udp_socket, public_endpoint) {
                     Ok((h, t)) => {
                         let _ = cmd_sender.send(Closure::new(move |state: &mut State| {
                             let _ = state.handle_rendezvous_connect(token, h, t);
                         }));
-                    },
+                    }
```
vinipsmaker: https://github.com/maidsafe/crust/pull/468/files#r49166927

src/service.rs
```diff
@@ -303,18 +307,18 @@ impl Service {
...
                     Err(e) => {
                         let _ = cmd_sender.send(Closure::new(move |state: &mut State| {
-                            let _ = state.event_sender.send(Event::OnRendezvousConnect(Err(e), token));
+                            let _ = state.event_sender
+                                         .send(Event::OnRendezvousConnect(Err(e), token));
                         }));
-                    },
+                    }
```
vinipsmaker: https://github.com/maidsafe/crust/pull/468/files#r49167202

vinipsmaker:
Forget my comments about commas on match expressions. I didn't know the rules were different if you used curly braces.

vinipsmaker:
Rest of PR seems okay.

src/service.rs
dirvine: rustfmt does this AFAIK

src/hole_punching.rs
```diff
@@ -220,12 +235,12 @@ impl HolePunchServer {
...
-                 * Currently we set a read timeout so that the hole punch server thread continually wakes
-                 * and checks to see if it's time to exit. This is a really crappy way of implementing
-                 * this but currently rust doesn't have a good cross-platform select/epoll interface.
-                 */
+        //
+        // TODO (canndrew):
+        // Currently we set a read timeout so that the hole punch server thread continually wakes
+        // and checks to see if it's time to exit. This is a really crappy way of implementing
+        // this but currently rust doesn't have a good cross-platform select/epoll interface.
+        //
```
inetic: Indendation seems odd here.

inetic: And also in few more places below

dirvine: Yes rustfmt does it this way for now

src/service.rs
```diff
@@ -214,12 +213,16 @@ impl Service {
...
 
                 match *port {
-                    Port::Tcp(_) => { let _ = TcpStream::connect(&addr); },
-                    Port::Utp(_) => { let _ = UtpSocket::connect(&addr); }
+                    Port::Tcp(_) => {
+                        let _ = TcpStream::connect(&addr);
+                    }
+                    Port::Utp(_) => {
+                        let _ = UtpSocket::connect(&addr);
+                    }
```
inetic: Personally I think it is easier to see the differences between the two match statements in the original.

dirvine: It's rustfmt doing this. If we feel very strongly about something in particular you can add a pragma to preserve it.We did this wen we had a docopts string in the routing ci_test

src/hole_punching.rs
inetic: I can understand the switch from '/* ... */' to just '//', but I don't understand the rule for indentation, why do we indent minus 8 spaces?

src/service.rs
inetic: fair enough

PR #471, opened 2016-01-08
Make Service::stop private

madadam:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/471)


maidsafe-highfive:
r? @vinipsmaker

PR #473, opened 2016-01-11
WIP fixing file handling

dirvine:
file handler needs rewrite, using lots of libc:: and very little rust to handle writing and reading files. Issue is serialisation, but will require work in HO as this part is in need of a refactor.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/473)


maidsafe-highfive:
r? @inetic

PR #477, opened 2016-01-12
WIP - continued cleanup

dirvine:
Removed Acceptor 
Use socket_addr::SocketAddr (serialisable) - still unsure of this one as perhaps it make smore sense to always pass endpoints to allow easy addition of more protocols etc. 
TODO
1. Compile 
2. Get tests passing again
3. Remove callbacks being sent on channels etc.
4. Agree v1.0.0 API
5. Implement API
   6 Create network wide tests or holepunching, udt etc. test via home pc's digital ocean etc. 

Nice 
Cut down threads considerably, potentially look at integrating mio as heart of system, possibly look at Andrews mioco (if we can be in stable)

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/477)


maidsafe-highfive:
r? @inetic

PR #478, opened 2016-01-13
Clean up endpoints etc

ustulation:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/478)


maidsafe-highfive:
r? @vinipsmaker

examples/crust_peer.rs
```diff
@@ -179,7 +174,7 @@ impl UdpData {
 
     pub fn send_to(&self, destination: SocketAddr) {
         let buf = [0u8; 256];
-        assert!(self.socket.send_to(&buf, destination).is_ok());
+        assert!(self.socket.send_to(&buf, &*destination).is_ok());
```
vinipsmaker: ?!?!?!

dirvine: We implemented `Deref` for our `socket_addr::socketAddr` to give a `std::net::ScoketAddr` so this is a dereferenced SocketAddr by reference :-( It is mostly because we wish to serialise SocketAddr and that was not implemented for net::SocketAddr, although the addition of having to deal with `ip::IpAddr` does add even more confusion to the code, just not specifically here.

PR #484, opened 2016-01-15
Fix 482

vinipsmaker:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/484)

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -133,6 +133,36 @@ pub fn is_local(ip_addr: &IpAddr, interfaces: &[Interface]) -> bool {
     false
 }
 
+fn ip(addr: &net::SocketAddr) -> IpAddr {
+    match *addr {
+        net::SocketAddr::V4(ref addr) => IpAddr::V4(*addr.ip()),
+        net::SocketAddr::V6(ref addr) => IpAddr::V6(*addr.ip()),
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @inetic

src/util.rs
```diff
@@ -133,6 +133,36 @@ pub fn is_local(ip_addr: &IpAddr, interfaces: &[Interface]) -> bool {
...
 
+fn ip(addr: &net::SocketAddr) -> IpAddr {
+    match *addr {
+        net::SocketAddr::V4(ref addr) => IpAddr::V4(*addr.ip()),
+        net::SocketAddr::V6(ref addr) => IpAddr::V6(*addr.ip()),
+    }
+}
+
+/// If the endpoint IP address is unspecified return a copy of the endpoint with the IP address
+/// set to the loopback address. Otherwise return a copy of the endpoint.
```
dirvine: This should be config test AFAIK as it's only used for the test function. All in all I feel it needs removed though.

vinipsmaker: https://github.com/maidsafe/crust/pull/484/files#diff-f7206c90a05dcea1bf538304ccae0949R109 is production code, not test code.

PR #489, opened 2016-01-19
Add OurContactInfo::make_their_info method

canndrew:
Routing will need this.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/489)

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -51,6 +51,18 @@ pub struct TheirContactInfo {
     pub rendezvous_addrs: Vec<SocketAddr>,
 }
 
+impl OurContactInfo {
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @vinipsmaker

vinipsmaker:
`OurContactInfo::make_their_info` isn't used nor part of the public API. Won't the compiler complain about unused method?

canndrew:
It should be part of the public.API :/. The compiler didn't complain.

vinipsmaker:
> It should be part of the public.API :/.

[My bad](http://www.urbandictionary.com/define.php?term=My+bad&defid=796865).

PR #493, opened 2016-01-20
Remove the obsolete OnHolePunched event.

afck:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/493)


afck:
I also exported `ContactInfoResult`, as that needs to be used to handle a `ContactInfoPrepared` event.

PR #508, opened 2016-01-29
Two services can talk to each other - test passes

ustulation:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/508)


maidsafe-highfive:
r? @vinipsmaker

PR #513, opened 2016-02-03
Service::connect and Service::prepare_connection_info are tested and working.

canndrew:
`Service::connect` and `Service::prepare_connection_info` are tested and working. This should be enough for routing to use and make utp connections.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/513)

### DIFF
--- a/src/bootstrap.rs
+++ b/src/bootstrap.rs
@@ -79,7 +79,7 @@ impl RaiiBootstrap {
                 break;
             }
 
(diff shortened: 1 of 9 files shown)


maidsafe-highfive:
r? @vinipsmaker

vinipsmaker:
Build fails on OSX. I tried restarting the build several times (given udp rendezvous not always succeed and we need to retry a few times), but it kept failing.

PR #519, opened 2016-02-09
Make Connection private.

afck:
Only pass the peer's keys to the user and handle the Connection objects
in Service.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/519)


maidsafe-highfive:
r? @canndrew

vinipsmaker:
As this commit is being introduced in [another PR](https://github.com/maidsafe/crust/pull/520), I'm closing it.

PR #526, opened 2016-02-12
Implement key exchange

canndrew:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/526)


maidsafe-highfive:
r? @vinipsmaker

PR #536, opened 2016-02-18
Service test changes

canndrew:
This adds another test to service.rs that tests connecting 3 peers via rendezvous connect and sending a bunch of data between them.
It also improves the UtpWrapper implementation to make it more efficient.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/536)


maidsafe-highfive:
r? @vinipsmaker

src/service.rs
```diff
@@ -397,16 +397,30 @@ impl Service {
...
 }
 
+impl Drop for Service {
+    fn drop(&mut self) {
+        // Disconnect from all peers when we drop the service
+        let mut cm = unwrap_result!(self.connection_map.lock());
+        cm.clear();
+    }
+}
+
```
vinipsmaker: This change was already introduced in this PR: https://github.com/maidsafe/crust/pull/537/files#diff-bfa17b8ec2b0aac77da5f96cbcbe4101R400

vinipsmaker:
your change needs Rust unstable: https://travis-ci.org/maidsafe/crust/jobs/110108351#L421

src/service.rs
```diff
@@ -430,6 +438,28 @@ mod test {
          event_rx)
     }
 
+    fn timebomb<R, F>(dur: Duration, f: F) -> R
```
afck: I'd find this a lot easier to read with an explaining comment, or maybe a more informative name. (`run_with_timeout`?)

src/service.rs
```diff
@@ -546,6 +576,27 @@ mod test {
             Event::LostPeer(id) => assert_eq!(id, id_0),
             e => panic!("Received unexpected event: {:?}", e),
         }
+
+        // Drop services and make sure the event channels close
+        drop(service_0);
+        let (done_tx, done_rx) = mpsc::channel();
+        let tj = thread!("Drain event channel messages", move || {
+            for _ in event_rx_0 {}
```
afck: Don't we know exactly at this point which remaining events to expect? (Probably none at all?)

src/service.rs
```diff
@@ -713,5 +764,173 @@ mod test {
...
+                        self.service.connect(our_ci, their_ci);
+                    }
+                    let mut their_ids = HashMap::new();
+                    for _ in 0..(NUM_SERVICES - 1) {
+                        let their_id = match unwrap_result!(self.event_rx.recv()) {
+                            Event::NewPeer(Ok(()), their_id) => their_id,
+                            m => panic!("Expected NewPeer message. Got message {:?}", m),
+                        };
+                        match their_ids.insert(their_id, 0u32) {
+                            Some(_) => panic!("Recieved two NewPeer events for same peer!"),
```
afck: "Received"

src/service.rs
```diff
@@ -713,5 +764,173 @@ mod test {
...
+                                match their_ids.entry(their_id.clone()) {
+                                    hash_map::Entry::Occupied(mut oe) => {
+                                        let next_msg = oe.get_mut();
+                                        assert_eq!(*next_msg, n as u32);
+                                        *next_msg += 1;
+                                    },
+                                    hash_map::Entry::Vacant(ve) => panic!("impossible!"),
+                                }
+                            },
+                            m => panic!("ERROR ERROR {} Unexpected msg receiving NewMessage: {:?}", ::time::SteadyTime::now(), m),
```
afck: With `use time::SteadyTime` and without the "ERROR ERROR" this would probably properly fit in the line.

src/utp_wrapper.rs
```diff
@@ -34,11 +38,16 @@ impl UtpWrapper {
                     socket.set_read_timeout(Some(CHECK_FOR_NEW_WRITES_INTERVAL_MS));
                     'outer: loop {
                         let mut buf = [0; BUFFER_SIZE];
-                        match socket.recv_from(&mut buf) {
+                        match socket.recv_from(&mut buf[..]) {
                             Ok((0, _src)) => break,
                             Ok((amt, _src)) => {
                                 let buf = &buf[..amt];
-                                let _ = input_tx.send(Vec::from(buf));
+                                match input_tx.send(Vec::from(buf)) {
```
afck: Clippy will probably demand `if let Err(mpsc::SendError(_)) =` here instead.

src/service.rs
canndrew: Ah, that's me leaving my test debugging in by accident.

src/utp_wrapper.rs
canndrew: That's a bit silly. I like to see exactly what data is getting thrown away. `if let` obscures that.

vinipsmaker:
It's failing to build on Rust stable: https://travis-ci.org/maidsafe/crust/jobs/111444771

vinipsmaker:
It's weird, but `start_three_service_udp_rendezvous_connect` is failing on stable Rust on CI machines. It works on my local machine.

canndrew:
I fixed a bug in `start_three_service_udp_rendezvous_connect`, but it's not the one that's causing those failures. Those failures seem to be because udp is currently disabled on master.

canndrew:
I've rebased this PR off of master and reenabled udp rendezvous connect. The current connect logic is:
  * Try connecting to all tcp endpoints in sequence and return if any are successful
  * Try to do udp rendezvous connect

Whether this is the best logic is certainly up for debate but for now at least we can connect over both protocols and all tests should pass.

vinipsmaker:
Reviewed 1 of 4 files at r1, 1 of 2 files at r3, 3 of 3 files at r7.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/536)*

PR #543, opened 2016-02-23
Send keepalive packet in utp multiplexer

canndrew:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/543)


maidsafe-highfive:
r? @vinipsmaker

src/lib.rs
```diff
@@ -61,10 +61,11 @@ extern crate get_if_addrs;
 extern crate sodiumoxide;
 extern crate config_file_handler;
 extern crate service_discovery;
-
 #[cfg(target_family = "unix")]
 extern crate libc;
 extern crate nat_traversal;
+#[cfg(test)]
+extern crate void;
```
Fraser999: Does this mean we can have void as a dev dependency only?

canndrew: It does! (Although it's a minuscule crate anyway)

src/utp_wrapper.rs
```diff
@@ -56,6 +67,9 @@ impl UtpWrapper {
                                         Err(TryRecvError::Empty) => break,
                                     }
                                 }
+                                if send_keepalive {
+                                    socket.send_keepalive();
```
vinipsmaker: This should be out of the write loop. If we only read for a long while, we will be silent and not send any keep alive.

vinipsmaker:
It needs to be rebased. It contains commits that were merged in master already.

src/utp_wrapper.rs
canndrew: It is outside of the write loop. The only thing inside the write loop is `send_keepalive = false` which tells us not to bother with the keepalive if we've been writing.

vinipsmaker: I got the identation wrong somehow. You're right.

Cargo.toml
```diff
@@ -31,7 +31,6 @@ sodiumoxide = "~0.0.9"
 temp_utp = "~0.7.0"
```
vinipsmaker: update to "~0.8.0"

Fraser999: We have a few deps listed here without `~` - could that be fixed too please?

vinipsmaker: > We have a few deps listed here without `~` - could that be fixed too please?

This sounds like work for another PR. I filled an issue and will work on that soon: https://github.com/maidsafe/crust/issues/550

canndrew: Eh, I already fixed it.

vinipsmaker:
Reviewed 2 of 4 files at r1, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/543)*

PR #548, opened 2016-02-25
Service test changes

vinipsmaker:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/548)


maidsafe-highfive:
r? @canndrew

canndrew:
We really shouldn't need an entire second per message on that test. Or even an entire 100ms for that matter.

PR #558, opened 2016-02-26
Fix notification order

vinipsmaker:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/558)

### DIFF
--- a/src/bootstrap.rs
+++ b/src/bootstrap.rs
@@ -99,23 +99,14 @@ impl RaiiBootstrap {
 
             // 1st try a TCP connect
             // 2nd try a UDP connection (and upgrade to UTP)
-            let connect_result = ::connection::connect(contact,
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @canndrew

canndrew:
In `utp_rendezvous_connect` I don't like that we block while holding the lock on the connection map. Other than that though this looks good.

vinipsmaker:
> In `utp_rendezvous_connect` I don't like that we block while holding the lock on the connection map. Other than that though this looks good.

It should be fixed now.

canndrew:
Reviewed 1 of 3 files at r1, 2 of 3 files at r2, 1 of 1 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/558)*

PR #570, opened 2016-03-03
Add stress test for sending/receiving messages

madadam:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/570)

### DIFF
--- a/tests/send_and_receive.rs
+++ b/tests/send_and_receive.rs
@@ -54,10 +54,7 @@ fn spawn_receiving_node(expected_connections: usize) -> JoinHandle<usize> {
     let mut service = unwrap_result!(Service::new(event_sender, BEACON_PORT));
 
     service.start_service_discovery();
-
-    // FIXME: tcp sometimes crashes during connect on osx.
-    // let _ = unwrap_result!(service.start_listening_tcp());
-
+    let _ = unwrap_result!(service.start_listening_tcp());
     let _ = unwrap_result!(service.start_listening_utp());
 
     // Wait for BootstrapFinished so we know this node is already listening when
--- a/tests/send_and_receive.rs
+++ b/tests/send_and_receive.rs
@@ -54,10 +54,11 @@ fn spawn_receiving_node(expected_connections: usize) -> JoinHandle<usize> {
     let mut service = unwrap_result!(Service::new(event_sender, BEACON_PORT));
 
     service.start_service_discovery();
-    let _ = unwrap_result!(service.start_listening_tcp());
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @vinipsmaker

vinipsmaker:
Test fails: https://travis-ci.org/maidsafe/crust/jobs/113421034#L930

If you can somehow put something like `#[ignore]` we could get this merged and somebody from the crust team could have a look to understand the failure.

madadam:
Yes, I'm planing to look into it sometime today, once I'm done with other stuff. If I don't figure it out in reasonable time, I'll `#[ignore]` it on mac only and move on.

madadam:
@vinipsmaker : It's all fixed now. Ready to review.

tests/send_and_receive.rs
```diff
@@ -0,0 +1,164 @@
...
+extern crate crust;
+#[macro_use]
+extern crate maidsafe_utilities;
+
+use crust::{CrustEventSender, Event, Service};
+use maidsafe_utilities::event_sender::{MaidSafeEventCategory, MaidSafeObserver};
+use std::sync::mpsc::{self, Receiver};
+use std::thread::{self, JoinHandle};
+
+const BEACON_PORT: u16 = 44444;
```
Fraser999: Using hard-coded values for ports in tests has previously caused problems when the port has been unavailable on the test machine (often because another test has acquired the port).  Is it OK here?

madadam: It's not ideal, but what is the alternative?

vinipsmaker: It could be if we start multiple tests in parallel. But it should be okay.

vinipsmaker:
Reviewed 2 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/570)*

tests/send_and_receive.rs
vinipsmaker: The problem is that we cannot just bind a socket to steal the port as it is UDP. So there is no way knowing somebody else isn't using the port already. As it is for service discovery, it should be less problematic. The test could still fail if they discover nodes from a different node also running service discovery.

The only option to really fix the issue here would be to use rendezvous connections and manually connect each node, but it is very unlikely we'll step upon the fault case. Plus, this "safety" would add to test robustness, not to crust robustness, so it is not that critical. And robustness this time means avoiding false negatives, so it is even more justifiable to accept this simpler code.

PR #573, opened 2016-03-04
Replace ip::IpAddr with std IpAddr

canndrew:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/573)


maidsafe-highfive:
r? @vinipsmaker

afck:
Merged into #575

PR #575, opened 2016-03-04
Remove ip dependency and replace cbor with maidsafe_utilities::serialisation.

afck:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/575)


maidsafe-highfive:
r? @vinipsmaker

PR #579, opened 2016-03-08
Tcp hole punching

canndrew:
Implement TCP hole punching.
This PR is based off of https://github.com/maidsafe/crust/pull/572

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/579)

PR #581, opened 2016-03-08
Tcp hole punching

vinipsmaker:
Continuing from https://github.com/maidsafe/crust/pull/579

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/crust/581)


maidsafe-highfive:
r? @canndrew

PR #592, opened 2016-03-15
Improve logging

canndrew:
Make the crust_peer example print logging information to the terminal rather than going through the maidsafe_utilities logging framework. Replace every occurrence of `WResult::result_discard` with `WResult::result_log` so that we can see the warnings that are raised.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/592)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ time = "~0.1.34"
 [dev-dependencies]
 docopt = "~0.6.78"
 void = "~1.0.1"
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @vinipsmaker

vinipsmaker:
Reviewed 5 of 5 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/592)*

Issue #588, opened 2016-03-11
Crust doesn't include acceptor ports in connection info.


afck:
I enabled port forwarding in my router and set that port as both `tcp_acceptor_port` and `utp_acceptor_port` in the config file. We also had working mapper servers enabled for both. However, only my local IP addresses showed up with the forwarded port in the connection info, but not the external one.

afck:
The problem seems to be that my router chooses a new external port for every outgoing connection, even if I forward 7000 to 7000 and use that port for sending data to the mapper server. So the mapper server never even gets the information that it should try connecting to me on port 7000.

This could probably either be solved in nat_traversal by telling the mapper server to try a specific port in addition to the one it got the message from (that would verify that port forwarding is configured correctly), or in Crust, by just always adding that port to the endpoints.

afck:
As discussed, we will go with the quick solution in Crust for now: Always add to the endpoints all our IP addresses together with the acceptor port.

canndrew:
r? @canndrew

I'll fix this tomorrow morning.

canndrew:
This should be fixed now @andreas, can you confirm that it works on your router?

afck:
I will verify it in April, when I'm behind that router again.

PR #593, opened 2016-03-15
Map tcp listener

canndrew:
Before, we were just getting our static tcp listening endpoints through `get_if_addrs`. This uses `MappedTcpSocket` instead. Also includes a fix for #588

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/593)


maidsafe-highfive:
r? @vinipsmaker

canndrew:
This requires https://github.com/maidsafe/nat_traversal/pull/58 before it will build.

vinipsmaker:
Reviewed 1 of 5 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.

---

*[src/connection.rs, line 433 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/593#-KD1ZHmBVdgjiwFHNdf0:-KD1ZHmDqfR3ZNG-1Zo-:-261013649) ([raw file](https://github.com/maidsafe/crust/blob/168a93ea00077720710081118fe2a45d9b9c11c4/src/connection.rs#L433)):*
Usually we avoid indexing patterns, but I see why you use it here and I think it's okay.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/593)*

PR #594, opened 2016-03-15
Lower TCP keepalive for testing

canndrew:
Set the keepalive to 10 seconds on all TCP connections as requested by @dirvine/@viv.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/594)

### DIFF
--- a/src/tcp_connections.rs
+++ b/src/tcp_connections.rs
@@ -21,8 +21,9 @@ use std::io;
 use std::thread;
 use std::sync::mpsc;
 use std::sync::mpsc::Sender;
+use std::time::Duration;
 use std::io::{Read, Write};
-use net2::TcpBuilder;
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @vinipsmaker

Viv-Rajkumar:
@canndrew this does not seem to help btw. With this PR had crust_peer run against two machines in the local network and when one was pulled off the network, the other peer did not detect the loss in connection.

canndrew:
Interesting. Do you know if it worked on OSX? On Linux and Windows I can diddle some of the other options to make the timeout even more aggressive but it doesn't look like it's possible on OSX.

vinipsmaker:
> but it doesn't look like it's possible on OSX.

I hope you find something so we do not need to resort to some more expensive ping-pong/keep-alive crust packets.

Issue #595, opened 2016-03-15
Bootstrap gets stuck if there is an invalid UDP contact in the crust config file or the bootstrap cache


madadam:
Steps to reproduce:

- create crust config file whose hard_coded_contacts field contains an invalid UDP contact (you can use 192.0.2.0, which is a documentation-only IP address which nobody should listen on.)
- create a service and make it use this config file
- wait for `Event::BootstrapFinished` from this service.
- you are going to wait forever :cry: 

The problem could possibly be fixed by setting a read timeout on the UDP socket.

canndrew:
@madadam was this fixed in #608?

vinipsmaker:
> @madadam was this fixed in #608?

I believe it was. If I'm not mistaken, this is the line changed that should fix the problem: https://github.com/maidsafe/crust/pull/608/files#diff-c80c453b374ae714374e0f9d9d86ba1cR193

madadam:
Yes, should be fixed. Closing.

PR #596, opened 2016-03-15
Fix bootstrap getting stuck on invalid UDP contact

madadam:
This is a quick and dirty fix for this #595.

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/596)


maidsafe-highfive:
r? @canndrew

madadam:
OK. I think this problem exists for TCP too, after all. Going to investigate.

madadam:
This will be bundled with the NewPeer event fix. Closing.

PR #602, opened 2016-03-17
More config options

madadam:
- Beacon port is now config option instead of constructor param
- Bootstrap cache filename is also config option
- Renamed `Service::new_with_config` to `Service::with_config` to be more in line with the convention in the std library (`Vec::new` vs `Vec::with_capacity`, ...).
- Added public const `DEFAULT_BEACON_PORT`
- rustfmt
- removed the `RUST_TEST_THREADS=1` workaround

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/602)


maidsafe-highfive:
r? @canndrew

canndrew:
This is now #609

PR #598, opened 2016-03-16
improve uTP bootstrap connect

vinipsmaker:
Fixes #587

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/598)

PR #610, opened 2016-03-18
improve uTP bootstrap connect

### LINKED ISSUE #587: Improve uTP connection logic
We should send the connection _syn_ to every address from the remote peer at once and only start waiting for a proper receive. Sending part should be done in parallel, so we don't risk receiving the message before we call the `recv_from` function (and thus losing the packet).

https://github.com/maidsafe/crust/blob/1d0392fc42ac95dce7c3c9af03e3744091aefb3c/src/connection.rs#L188

canndrew:
Fixes #587 
This is a rebase of #598

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/610)


maidsafe-highfive:
r? @vinipsmaker

Issue #600, opened 2016-03-16
Add option to either bypass bootstrap cache or override its location


madadam:
The bootstrap cache interferes with multithreaded testing: Sometimes a service in thread A creates the bootstrap cache and service in thread B reads it and tries to bootstrap against it. B then ends up bootstrapping against A, which causes non-deterministic test failures.

Current workaround: force `RUST_TEST_THREADS=1`. Not nice as it makes the test slower.
Possible fix: Add option to disable bootstrap cache (at least in tests)
Another possible fix: Add option to override the location of the cache file. Then each test could set it to a unique, temporary file.

canndrew:
Fixed in #609

PR #609, opened 2016-03-18
More config options

canndrew:
This is a rebase of #602 against master.

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/609)

PR #611, opened 2016-03-18
Thread-Local bootstrap cache filename

ligthyear:
Replaces bootstrap_handlers' get_file_name in test with a thread-local
version. This local one returns a filename for the current thread only
ensuring we can't run into race bootstrap-file-caused race conditions.

Comes with test that ensure we are getting different file names during
testing. (Not sure why the other tests in there are commented out
though.)

Fixes maidsafe/crust#600

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/611)

### DIFF
--- a/src/bootstrap_handler.rs
+++ b/src/bootstrap_handler.rs
@@ -99,6 +99,17 @@ impl BootstrapHandler {
     }
 }
 
+#[cfg(test)]
+#[allow(unsafe_code)]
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:

canndrew:
Hi, thanks for the PR! I'm gonna close it though because this just got fixed a different way in #609. We now allow the bootstrap cache filename to be passed as a config option and we use that during tests.

PR #612, opened 2016-03-18
Update sample.config for config changes

canndrew:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/612)


maidsafe-highfive:
r? @vinipsmaker

PR #613, opened 2016-03-18
maidsafe_utilities version update for bincode

canndrew:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/613)


maidsafe-highfive:
r? @vinipsmaker

PR #616, opened 2016-03-21
Not raise `NewPeer`/`LostPeer` for ourselves

vinipsmaker:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/616)

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -221,6 +221,10 @@ pub fn connect(peer_contact: StaticContactInfo,
                     match deserialise::<ListenerResponse>(&read_buf[..bytes_rxd]) {
                         Ok(ListenerResponse::Connect { our_info, their_info, pub_key, }) => {
                             let (their_info, our_info) = (our_info, their_info);
+                            let their_pub_key = pub_key;
+                            if our_public_key == their_pub_key {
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @canndrew

src/connection.rs
```diff
@@ -221,6 +221,10 @@ pub fn connect(peer_contact: StaticContactInfo,
                     match deserialise::<ListenerResponse>(&read_buf[..bytes_rxd]) {
                         Ok(ListenerResponse::Connect { our_info, their_info, pub_key, }) => {
                             let (their_info, our_info) = (our_info, their_info);
+                            let their_pub_key = pub_key;
```
madadam: Nitpick, but can you instead rename the `pub_key` variable on line 222 to `their_pub_key` and then you wouldn't need this line at all?

vinipsmaker: I don't like the idea of renaming the variable on the struct. This struct is serialized and sent over the wire. It has different meanings when it's sent and when it's received.

madadam: Sorry, you are right. For some reason I though that it's a tuple, not struct. My bad.

PR #624, opened 2016-03-23
Remove connect_utp, reenable unused lint

canndrew:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/624)

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -25,6 +25,7 @@ use std::net::{Shutdown, TcpStream, UdpSocket, Ipv4Addr, SocketAddrV4};
 use std::net;
 use std::io;
 use std::time::Duration;
+use time;
 use maidsafe_utilities::event_sender::{EventSenderError, MaidSafeEventCategory};
 use maidsafe_utilities::thread::RaiiThreadJoiner;
 use maidsafe_utilities::serialisation::{deserialise, serialise};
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @vinipsmaker

src/connection.rs
```diff
@@ -248,6 +253,15 @@ pub fn connect(peer_contact: StaticContactInfo,
     Err(last_err)
 }
 
+/*  FIXME(canndrew) 23/03/2016
```
vinipsmaker: you can put a `#[allow(unused)]`/`#[warn(unused)]` just before the function that should ignore the lint instead commenting. But I'm not sure about the trade-off (compilation time of unused function or maintainability of not breaking commented code accidentally). As it'll be enabled very soon, I don't mind about keeping the code commented.

src/utp_connections.rs
```diff
@@ -65,59 +60,19 @@ mod test {
...
-        let _ = handle.join().unwrap();
-    }
+        let port_0 = unwrap_result!(socket_0.local_addr()).port();
+        let port_1 = unwrap_result!(socket_1.local_addr()).port();
 
-    #[test]
-    fn send_receive_data() {
-        let listener = listen(0).unwrap();
-        let port = listener.local_addr().unwrap().port();
+        let addr_0 = SocketAddr(net::SocketAddr::V4(net::SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), port_0)));
```
vinipsmaker: why don't we use that function here, `unspecified_to_localhost`?

canndrew: It would make more sense to add a new function like

```
fn SocketAddr::localhost_v4(port: u16) -> SocketAddr;
```

Actually I think that should be added to the standard library. I might raise a rust-lang/rfcs bug about it.

vinipsmaker:
Rebased as https://github.com/maidsafe/crust/pull/627

PR #629, opened 2016-03-25
Fixes

vinipsmaker:
r? @afck

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/629)

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -44,6 +44,7 @@ use bootstrap_handler::BootstrapHandler;
 use nat_traversal::{MappedUdpSocket, MappingContext, PunchedUdpSocket, gen_rendezvous_info};
 use nat_traversal;
 use sodiumoxide::crypto::box_::PublicKey;
+use util::time_duration_to_std_duration;
 
 type CrustEventSenderError = EventSenderError<MaidSafeEventCategory, Event>;
 
@@ -202,7 +203,7 @@ pub fn connect(peer_contact: StaticContactInfo,
         loop {
             let now = time::SteadyTime::now();
             let timeout = if now < deadline { deadline - now } else { break };
-            let timeout = Duration::from_millis(timeout.num_milliseconds() as u64);
+            let timeout = time_duration_to_std_duration(timeout);
             try!(udp_socket.set_read_timeout(Some(timeout)));
             match udp_socket.recv_from(&mut read_buf) {
                 Ok((bytes_rxd, _peer_addr)) => {
--- a/src/service.rs
+++ b/src/service.rs
@@ -812,7 +812,7 @@ mod test {
             };
             unwrap_result!(done_tx.send(()));
         });
-        thread::park_timeout(Duration::from_secs(5));
+        thread::park_timeout(Duration::from_secs(60));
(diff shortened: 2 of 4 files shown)


afck:

---

*[src/connection.rs, line 209 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/629#-KDieuVFdI4T2nI8zS8f:-KDieuVGyj3-cPLwX4BA:-1044363147) ([raw file](https://github.com/maidsafe/crust/blob/e07e54947c302ae9019d96edc8ea29f63b64b912/src/connection.rs#L209)):*
These lines should not be needed if the conversion function would work correctly, since `deadline > now`. (See below.)

---

*[src/util.rs, line 28 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/629#-KDifASgMBKo7uhzH3nv:-KDifAShopnv_6UHs9BC:1542146510) ([raw file](https://github.com/maidsafe/crust/blob/e07e54947c302ae9019d96edc8ea29f63b64b912/src/util.rs#L28)):*
If you match `(dur - Duration::seconds(secs)).num_nanoseconds()` instead of subtracting that inside, this will never be `None` and the result will be exact to the nanosecond.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/629)*

afck:

---

*[src/util.rs, line 28 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/629#-KDifASgMBKo7uhzH3nv:-KDiiv6tdNWwGZC_j2tl:-1548947178) ([raw file](https://github.com/maidsafe/crust/blob/e07e54947c302ae9019d96edc8ea29f63b64b912/src/util.rs#L28)):*
I wouldn't even do the checks for `< 0` now anymore. Without them, it should just always produce the `Duration` with the exact same two values for `seconds` and `nanoseconds`, regardless of whether they are positive or negative.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/629)*

afck:
Reviewed 2 of 4 files at r1, 1 of 2 files at r2, 1 of 1 files at r4.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/629)*

vinipsmaker:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/connection.rs, line 209 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/629#-KDieuVFdI4T2nI8zS8f:-KDxRBcSbKIDMgCR5TSU:-499146009) ([raw file](https://github.com/maidsafe/crust/blob/e07e54947c302ae9019d96edc8ea29f63b64b912/src/connection.rs#L209)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/629)*

afck:
Reviewed 1 of 1 files at r5.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/629)*

PR #636, opened 2016-03-31
uTP listener fix

canndrew:
Make the utp listener send the `ListenerResponse` from the listening socket.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/636)

### DIFF
--- a/src/udp_listener.rs
+++ b/src/udp_listener.rs
@@ -27,6 +27,7 @@ use maidsafe_utilities::serialisation::{deserialise, serialise};
 use maidsafe_utilities::thread::RaiiThreadJoiner;
 use nat_traversal::{MappedUdpSocket, MappingContext, PunchedUdpSocket, gen_rendezvous_info};
 use sodiumoxide::crypto::box_::PublicKey;
+use crossbeam;
 
 use connection::{Connection, utp_rendezvous_connect, UtpRendezvousConnectMode};
 use static_contact_info::StaticContactInfo;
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @vinipsmaker

src/udp_listener.rs
```diff
@@ -94,26 +95,30 @@ impl RaiiUdpListener {
...
-                        RaiiUdpListener::handle_request(msg,
-                                                        our_public_key,
-                                                        peer_addr,
-                                                        event_tx,
-                                                        connection_map,
-                                                        mc);
-                    });
+
+        let udp_socket = &udp_socket;
+        crossbeam::scope(move |scope| {
```
vinipsmaker: I'm not sure I'm happy in creating yet another thread just to do a `udp_socket.clone()`. What do you think about [this alternative approach](https://github.com/maidsafe/crust/pull/638)?

canndrew:
`crossbeam::scoped` doesn't create a thread. It lets you use the `scope` to spawn threads that can borrow data from outside the call to scope (in this case `&udp_socket`),

vinipsmaker:
> `crossbeam::scoped` doesn't create a thread. It lets you use the `scope` to spawn threads that can borrow data from outside the call to scope (in this case `&udp_socket`),

Could we leave the loop of the udp listener thread and destroy the socket to only then let the second thread use the (destroyed) socket?

canndrew:
Nope, the purpose of `crossbeam::scope` is to allow you to safely share borrowed data between threads without overhead and without having to use `Arc` or `try_clone` or anything like that.

All it does it call the closure that it's given then block until any threads that were created inside of it (using the `Scope`) have exited. That way, the spawned threads can be tied to the lifetime of the `Scope` instead of having to be `'static`.

It's basically safe hack to relax the borrow checker.

vinipsmaker:
Still... is there any reason why you avoided `udp_socket.try_clone()`? It'd be much simpler.

canndrew:
To me using `crossbeam` seems simpler. Rather than duplicating a file descriptor (and having to handle the case that it might fail) just tell the compiler that it's fine if both threads use the same socket.

vinipsmaker:
You nested much more code just to do a `udp_socket.clone()`. It's not simpler code, but the main problem is the lack of uniformity. Why did you give different treatment for `udp_socket` and `mapping_context`. To me it looks like we taking different approaches to implement as many techniques as we can just to make the code more confuse.

canndrew:
> You nested much more code just to do a `udp_socket.clone()`. It's not simpler code,

Except I didn't do a clone. There's one reference to a socket and both threads are using it. This is conceptually simpler than having to clone a socket, it's exactly like using the same pointer from two threads in C/C++. It's just a side-effect of rust's borrow checker that you need to indent an extra block to do it.

> but the main problem is the lack of uniformity. Why did you give different treatment for `udp_socket` and `mapping_context`.

That's a good point, the other variables could be moved to outside the scope aswell so they don't have to be cloned (except for `event_tx` because it's not `Sync`). However cloning an `Arc` isn't the same as cloning a `UdpSocket`. Cloning an `Arc` just increments a reference count and always succeeds whereas cloning a `UdpSocket` does a sytem call which creates a new file descriptor and can fail.

If you feel strongly about it I'll change it but I disagree it's more complicated. Our current approach of using `Arc` everywhere is slow, bug-prone and risks leaking threads. We should be using scopes and sharing stack data between threads whereever possible.

vinipsmaker:
>> You nested much more code just to do a `udp_socket.clone()`. It's not simpler code,

> Except I didn't do a clone. There's one reference to a socket and both threads are using it. This is conceptually simpler than having to clone a socket, it's exactly like using the same pointer from two threads in C/C++. It's just a side-effect of rust's borrow checker that you need to indent an extra block to do it.

You already needed to understand why you're cloing because you were cloning other members. Your approach incurs more concepts. I still disagree with you on this one. I do believe it's more complex.

> If you feel strongly about it I'll change it but I disagree it's more complicated.

Never mind. Your approach saves one file descriptor. We had problems allocating them on OSX in the past because OSX have rather low limits. I don't mind complexity if it spares resources.

vinipsmaker:
Can you rebase your branch against master (remember to ping me back as GitHub won't notify me of new commits added to PR)?

vinipsmaker:
Reviewed 2 of 2 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/crust/636)*

PR #642, opened 2016-04-07
Fix tests

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/642)


maidsafe-highfive:
r? @vinipsmaker

PR #645, opened 2016-04-13
Heartbeat

vinipsmaker:
---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/645)

### DIFF
--- a/src/utp_wrapper.rs
+++ b/src/utp_wrapper.rs
@@ -9,6 +9,8 @@ use std::collections::VecDeque;
 use maidsafe_utilities::thread::RaiiThreadJoiner;
 use maidsafe_utilities::serialisation::serialise;
 use event::WriteEvent;
+use std::time::{Duration,Instant};
+use sender_receiver::CrustMsg;
 
 const CHECK_FOR_NEW_WRITES_INTERVAL_MS: u64 = 50;
 const BUFFER_SIZE: usize = 1000;
@@ -22,7 +24,10 @@ pub struct UtpWrapper {
 }
 
 impl UtpWrapper {
-    pub fn wrap(socket: UtpSocket, output_rx: Receiver<WriteEvent>) -> io::Result<UtpWrapper> {
+    pub fn wrap(socket: UtpSocket, output_rx: Receiver<WriteEvent>,
+                heartbeat_timeout: Duration,
+                inactivity_timeout: Duration)
+                -> io::Result<UtpWrapper> {
         let (input_tx, input_rx) = mpsc::channel();
         let peer_addr = try!(socket.peer_addr());
         let local_addr = try!(socket.local_addr());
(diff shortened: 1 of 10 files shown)


maidsafe-highfive:
r? @canndrew

vinipsmaker:
@canndrew: PR is ready for review.

src/utp_wrapper.rs
```diff
@@ -32,15 +37,40 @@ impl UtpWrapper {
...
+                                info!("Stale connection. Dropping...");
+                                break;
+                            }
+                            if now > heartbeat_deadline {
+                                if let Err(e) = socket.send_to(&heartbeat_msg) {
+                                    error!("Error sending: {:?}", e);
+                                    break;
+                                }
+                                update_heartbeat(&mut heartbeat_deadline,
+                                                 &mut inactivity_deadline);
```
canndrew: `inactivity_deadline` will get updated every time we send a heartbeat. So is it supposed to ever expire?

vinipsmaker: The `inactivity_timeout` is supposed to make us ignore the underlying protocol rules and drop the connection if we don't ever detect activity within the time given.

I just end up modelling this behaviour, but as you noticed, this timeout will only be reached when thread goes to sleep for a long time. Maybe I should just remove this variable. What do you think?

I also update the `inactivity_timeout` when a message is sent because this will also trigger the uTP socket to exchange messages with the remote peer.

canndrew: I don't think it needs to be gotten rid of. I think it's good to have it there. But the implementation should only update `inactivity_deadline` when it receives data from the peer, not whenever it sends a heartbeat. Also the `heartbeat_deadline` shouldn't be updated every time we receive data from the peer otherwise, if they're sending us data, we'll never send them a heartbeat and they may end up thinking we've disconnected. Ironically, this is the exact same bug that I left in the `rust-utp` implementation.

canndrew:
However a bigger issue here is the tcp implementation. I'm not really a fan of having the writer thread tick over every 10ms like that. Especially since, with the transport library, it won't be necessary to have a writer thread at all (because the transport library supplies its own writer thread). I think a better solution might be to have `Service` spawn a single thread that periodically writes a heartbeat to all connections in the connection map that haven't had any activity recently.

src/utp_wrapper.rs
vinipsmaker: > But the implementation should only update `inactivity_deadline` when it receives data from the peer, not whenever it sends a heartbeat.

But the process of sending hearbeats will force both peers to exchange messages (ack and stuff). Sending messages is blocking and signalizes errors through the return value, so I assume both peers communicated after `send` returns.

> Also the `heartbeat_deadline` shouldn't be updated every time we receive data from the peer otherwise, if they're sending us data, we'll never send them a heartbeat and they may end up thinking we've disconnected.

But we'll send them an ACK instead a heartbeat. We acknowledge every data we receive.

vinipsmaker:
> I think a better solution might be to have Service spawn a single thread that periodically writes a heartbeat to all connections in the connection map that haven't had any activity recently.

I'm not sure what's more expensive. I think I'll just write the alternative solution and then we can have a better idea. What do you think about this plan?

> Of course that would cover the utp case aswell :/

uTP case seems well implemented to me. Any reason why you think it should change too?

src/utp_wrapper.rs
canndrew: > so I assume both peers communicated after `send` returns.

Is that actually the case though? What happens if one peers has dropped offline and isn't sending ACKs back? I know with tcp you can keep sending data for a long time without getting acknowledged before the OS decides that the connection has died.

canndrew:
> uTP case seems well implemented to me. Any reason why you think it should change too?

I just meant that if we went with the single-hearbeat-thread model then utp wouldn't need to implement heartbeat itself.

src/utp_wrapper.rs
vinipsmaker: Fair enough. I'll change the TCP side only consider that `last_read_activity`.

Some other issue you might think of?

canndrew: Nah, that sounds good for now.

vinipsmaker: done: https://github.com/maidsafe/crust/pull/645/commits/c6b6074a164408e15e5f0ff99e6ebd517d400fd8

I should squash the two last commits into one, but I wanted to kept'em separate for now to ease your review (now you know what changed easily).

canndrew:
Reviewed 7 of 9 files at r2, 2 of 3 files at r3, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/645)*

PR #654, opened 2016-04-25
fix/tcp: heartbeat not being sent on some cases

vinipsmaker:
There was a scenario where hearbeat implementation would fail:

> if peer A is sending data to peer B and never receives a heartbeat
> it'll eventually conclude that peer B has disconnected (even if it was
> still getting acks at the tcp level).

To fix the problem, we should only take `last_write_activity` into
consideration to compute the heartbeat timeout.

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/654)

### DIFF
--- a/src/tcp_connections.rs
+++ b/src/tcp_connections.rs
@@ -88,12 +88,7 @@ fn upgrade_writer(mut stream: TcpStream,
                                                info!("Stale connection. Dropping...");
                                                break;
                                            }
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @canndrew

ustulation:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/654)*

PR #662, opened 2016-04-28
refactor/tcp: write payload size before payload

ustulation:
Fit to be merged

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/662)

### DIFF
--- a/src/tcp_connections.rs
+++ b/src/tcp_connections.rs
@@ -23,7 +23,7 @@ use std::sync::mpsc::Sender;
 use std::io::Write;
 
 use event::WriteEvent;
-use maidsafe_utilities::serialisation::serialise;
+// use maidsafe_utilities::serialisation::serialise;
 
 /// Connect to a peer and open a send-receive pair.  See `upgrade` for more details.
 pub fn connect_tcp(addr: SocketAddr) -> io::Result<(TcpStream, Sender<WriteEvent>)> {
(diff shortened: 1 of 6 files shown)


maidsafe-highfive:
r? @vinipsmaker

src/tcp_connections.rs
```diff
@@ -46,31 +45,43 @@ pub fn upgrade_tcp(stream: TcpStream) -> io::Result<(TcpStream, Sender<WriteEven
...
+    let _ = thread!("TCP writer", move || {
+        while let Ok(event) = rx.recv() {
+            match event {
+                WriteEvent::Write(data) => {
+                    use std::io::Write;
+                    use byteorder::{WriteBytesExt, LittleEndian};
+
+                    let payload = unwrap_result!(serialise_with_limit(&data, SizeLimit::Infinite));
+                    let size = payload.len() as u32;
+                    let mut little_endian_size_bytes = Vec::with_capacity(4);
```
canndrew: You could just use a `[u8; 4]` here.

ustulation: Ya that would be nice - i had used that initially, but seems that the trait is not implemented for `[u8; 4]`, so the next thing to do was to use a vec. Array will give a compile error:
```src/tcp_connections.rs:67:61: 67:70 error: no method named `write_u32` found for type `[u8; 4]` in the current scope
src/tcp_connections.rs:67                     unwrap_result!(little_endian_size_bytes.write_u32::<LittleEndian>(size));
                                                                                      ^~~~~~~~~
src/tcp_connections.rs:67:21: 67:94 note: in this expansion of unwrap_result! (defined in <maidsafe_utilities macros>)
<maidsafe_utilities macros>:2:1: 4:28 note: in this expansion of unwrap_result! (defined in <maidsafe_utilities macros>)
src/tcp_connections.rs:57:13: 82:7 note: in this expansion of thread! (defined in <maidsafe_utilities macros>)
src/tcp_connections.rs:67:61: 67:70 note: the method `write_u32` exists but the following trait bounds were not satisfied: `[u8; 4] : std::io::Write`, `[u8] : std::io::Write`
error: aborting due to previous error
Could not compile `crust`.
```

PR #669, opened 2016-05-06
feat/tcp_connection: log the number of connections, change heartbeat period

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/669)


maidsafe-highfive:
r? @canndrew

PR #681, opened 2016-05-12
feat/service: add network_name config option

afck:
Only nodes with the same network name will connect, so that separate test networks can be set up and incompatible versions can be prevented from interfering with each other.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/681)

### DIFF
--- a/installer/sample.config
+++ b/installer/sample.config
@@ -26,6 +26,7 @@
   "enable_tcp": true,
   "enable_utp": true,
   "service_discovery_port": null,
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @vinipsmaker

Viv-Rajkumar:
Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/681)*

PR #682, opened 2016-05-12
feat/service: migrate to nat_traversal 0.4.1 and set timeout to 60s

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/682)


maidsafe-highfive:
r? @vinipsmaker

PR #686, opened 2016-05-13
Async

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/686)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,16 @@
+*.exe
+*.lock
+*.swp
+*.rsproj
+tags*
+build/
+build-tests/
+target/
+*~
+/.idea
+tests/tmp*
+src/tmp*
+/.project
+*.sublime-*
+*.bootstrap.cache
+/bin/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,0 +1,52 @@
+[package]
+authors = ["MaidSafe Developers <dev@maidsafe.net>"]
+description = "Peer-to-peer networking library. Automatically reconnect and manage connections."
+documentation = "http://maidsafe.net/crust/latest"
+homepage = "http://maidsafe.net"
+license = "GPL-3.0"
+name = "crust"
+readme = "README.md"
+repository = "https://github.com/maidsafe/crust"
+version = "0.12.0"
+
+[dependencies]
+maidsafe_utilities = "~0.5.4"
+mio = "~0.5.1"
+quick-error = "~1.0.0"
+
+[dependencies.clippy]
+optional = true
+version = "~0.0.64"
+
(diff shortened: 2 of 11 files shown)


maidsafe-highfive:
r? @vinipsmaker

ustulation:
Review status: 0 of 10 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/service.rs, line 86 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/686#-KHdi5UdCN0UZERlVEEX:-KHdi5Uepj6BiZhvB-la:985371128) ([raw file](https://github.com/maidsafe/crust/blob/e65a04615842673f863e2993a4f29eecc3a785d4/src/service.rs#L86)):*
> ```Rust
> /// A structure representing a connection manager.
> pub struct Service {
>     cm: Arc<Mutex<HashMap<PeerId, Context>>>, // This is the connection map -> PeerId <-> Context
> ```

better to name this as conn_map - cm might be too short :)
Or still better have `pub type ConnectionMap = Arc<Mutex<HashMap<PeerId, Context>>` defined then we can use `cm` throughout.

---

*[src/service.rs, line 101 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/686#-KHdjRUUrHBuPE48jfW-:-KHdjRUUrHBuPE48jfW0:-878199252) ([raw file](https://github.com/maidsafe/crust/blob/e65a04615842673f863e2993a4f29eecc3a785d4/src/service.rs#L101)):*
> ```Rust
>         let mut event_loop = try!(EventLoop::new());
>         let mio_tx = event_loop.channel();
>         // let sender = event_loop.channel();
> ```

Let's delete them if not necessary

---

*[src/service.rs, line 221 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/686#-KHdjf9usfwYVc2hJ4Ou:-KHdjf9vyXG4IFcJR7r7:-1286796529) ([raw file](https://github.com/maidsafe/crust/blob/e65a04615842673f863e2993a4f29eecc3a785d4/src/service.rs#L221)):*
> ```Rust
> impl Drop for Service {
>     fn drop(&mut self) {
>         // let _ = self.sender.send(Box::new(|el: &mut EventLoop<Core>, _: &mut Core| el.shutdown()));
> ```

delete ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/686)*

ustulation:
Reviewed 3 of 9 files at r1, 6 of 8 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/686)*

PR #688, opened 2016-05-16
fix/connection_states: simple_demo exception and hanging issue on Windows

maqi:
Windows doesn't allow using different token to registration to the same socket.
Windows doesn't suppot socket try_clone

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/688)


maidsafe-highfive:
r? @vinipsmaker

PR #690, opened 2016-05-16
feat/active_connections: Send data_len first then data

maqi:
Sending data_len first then data.
On receiving side, read size first, then reading data

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/690)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/maidsafe/crust"
 version = "0.12.0"
 
 [dependencies]
+byteorder = "~0.5.1"
 maidsafe_utilities = "~0.5.4"
 mio = "~0.5.1"
 nat_traversal = "~0.3.3"
--- a/examples/simple_demo.rs
+++ b/examples/simple_demo.rs
@@ -44,19 +44,29 @@ use std::time::Duration;
 use crust::{CrustEventSender, Event, Service};
 use maidsafe_utilities::event_sender::MaidSafeEventCategory;
 use maidsafe_utilities::thread::RaiiThreadJoiner;
+use rand::Rng;
 
 fn spawn_test_server() {
     let ls = TcpListener::bind("127.0.0.1:33333").expect("Could not bind listener.");
     let (mut strm, _) = ls.accept().expect("Error in accepting");
     loop {
-        let mut buf = [0; 10];
+        let mut buf = [0; 14];
         if strm.read_exact(&mut buf).is_err() {
             // Graceful exit in this case
             break;
         }
         println!("Test peer received {:?}", buf);
-        strm.write_all(&mut buf).expect("Error in writing");
+        let mut reply : Vec<u8> = vec![0x04, 0x00, 0x00, 0x00,
+                                       0x11, 0x22, 0x33, 0x44,
+                                       0x04, 0x00, 0x00, 0x00,
+                                       0xAA, 0xBB, 0xCC, 0xDD];
+        strm.write_all(&mut reply).expect("Error in writing");
     }
+}
+
+/// utility to create random vec u8 of a given size
+pub fn generate_random_vec_u8(size: usize) -> Vec<u8> {
+    rand::thread_rng().gen_iter().take(size).collect()
 }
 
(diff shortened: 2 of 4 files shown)


maidsafe-highfive:
r? @vinipsmaker

ustulation:

---

*[src/connection_states/active_connection.rs, line 102 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/690#-KHu_6Hk85W_ZZUwozgl:-KHu_6Hk85W_ZZUwozgm:-320745628) ([raw file](https://github.com/maidsafe/crust/blob/1943a6e8b405f25bccb01d8619224c8f6cf8e37a/src/connection_states/active_connection.rs#L102)):*
> ```Rust
>     fn read_data(&mut self) {
>         if self.read_len == 0 && self.read_buf.len() >= U32_BYTE_LENGTH {
>             if let Ok(size) = Cursor::new(&self.read_buf[..]).read_u32::<LittleEndian>() {
> ```

Is it not safer to write this as `Cursor::new(&self.read_buf[..U32_BYTE_LENGTH]).read_u32::<LittleEndian>()`? Or is read_u32 immune to such errors ? Also if this fails, it's probably worth informing routing and closing connection because again the internal state of tracking the payload size and payload will be (probably) irrecoverably damaged. So maybe do a `.expect(...)` instead of `if let Ok(...)` here and put a TODO saying inform this to routing and gracefully terminate connection in actual production code

---

*[src/connection_states/active_connection.rs, line 112 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/690#-KHua7e_ctYUzPbZJ4bM:-KHua7e_ctYUzPbZJ4bN:1189455344) ([raw file](https://github.com/maidsafe/crust/blob/1943a6e8b405f25bccb01d8619224c8f6cf8e37a/src/connection_states/active_connection.rs#L112)):*
> ```Rust
>             self.read_buf = tail;
>             self.read_len = 0;
>             self.read_data();
> ```

Recursion may potentially leave an attack vector open - it's actually very easy to simulate - the sender can send great many number of small messages, say 20 MB of total data with each message as 20 bytes. This will cause huge number of recursions and it will be easier to get the receiver node panic with `Stack-Overflow`. Do you think a loop is much better here than recursion ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/690)*

maqi:
*[src/connection_states/active_connection.rs, line 102 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/690#-KHu_6Hk85W_ZZUwozgl:-KHw8rzOm0wgVcxFQNUp:806214532) ([raw file](https://github.com/maidsafe/crust/blob/1943a6e8b405f25bccb01d8619224c8f6cf8e37a/src/connection_states/active_connection.rs#L102)):*
> Is it not safer to write this as `Cursor::new(&self.read_buf[..U32_BYTE_LENGTH]).read_u32::<LittleEndian>()`? Or is read_u32 immune to such errors ? Also if this fails, it's probably worth informing routing and closing connection because again the internal state of tracking the payload size and payload will be (probably) irrecoverably damaged. So maybe do a `.expect(...)` instead of `if let Ok(...)` here and put a TODO saying inform this to routing and gracefully terminate connection in actual production code</details>
the simple_demo already shows `read_u32` can parse u32 from large vec.
However, I put in `[..U32_BYTE_LENGTH]` anyway.

I have put in `expect` and `TODO` comment as you mentioned. However, don't think it will happen, as `u32` use all 4 bytes 0x00 - 0xFF , so as long as their is 4 bytes, there will have a u32.

if the read size is incorrect, if less than real, then it will early return. if larger than real, it will hang. 
I think introducing time_out to prevent such hang shall be considered but be in a different PR? so I put a `TODO` commnet as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/690)*

ustulation:
Reviewed 3 of 4 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/690)*

PR #692, opened 2016-05-17
feat/tcp_connection: drop higher priorities on insufficient bandwidth

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/692)


maidsafe-highfive:
r? @canndrew

PR #699, opened 2016-05-20
fix/service_discovery: fix incorrect codes

ustulation:
1> put service discovery in proper module instead of where it was 2> code reduction 3> rectify incorrect usage of token in service.rs for service discovery - existing code could easily become buggy 4> proper hadling of interleaved reads and writes 5> code was logically buggy in set listen for service discovery 5> Test case becomes buggy without delay between initiation of service and setting service discovery to listen mode - corrected that too

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/699)


maidsafe-highfive:
r? @vinipsmaker

dirvine:
Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/699)*

PR #702, opened 2016-05-23
Nat

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/702)


maidsafe-highfive:
r? @canndrew

PR #705, opened 2016-05-24
Flush socket on write automatically

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/705)

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -110,7 +110,14 @@ impl Socket {
     }
 
     // Write a message to the socket.
-    pub fn write<T: Encodable>(&mut self, message: T) -> Result<(), SocketError> {
+    //
+    // Returns:
+    //   - Ok(true):   the message has been successfuly written.
+    //   - Ok(false):  the message (or part of it) has been queued. Schedule a
+    //                 write event and call `flush` to finish the write in the
+    //                 next ready handler.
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:
r? @canndrew

src/socket.rs
```diff
@@ -138,30 +144,28 @@ impl Socket {
     //                 the next invocation of the ready handler.
     //   - Err(error): there was an error while writing to the socket.
     pub fn flush(&mut self) -> Result<bool, SocketError> {
-        if let Some(mut data) = self.write_queue.pop_front() {
+        while let Some(mut data) = self.write_queue.pop_front() {
```
ustulation: The previous `if` is what is ideal here. In case the messages accumulate and then later when socket is writable we still dont want to write all messages in one go. Infact we dont even want to write single msg in one go. Only as much as socket.write(..) writes in one go. So this should be interleaved. Pop front .. write .. if all went then fine else push the remaining part of msg push front .. check if queue is empty .. if not ok(false) else Ok(true). That way we keep everything interleaved.

ustulation:
Reviewed 4 of 4 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/705)*

PR #707, opened 2016-05-24
Use service discovery in bootstrap

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/707)

### DIFF
--- a/src/bootstrap_states/establish_connection.rs
+++ b/src/bootstrap_states/establish_connection.rs
@@ -67,9 +67,19 @@ impl EstablishConnection {
             }
         };
 
+        let event_set = match socket.write(Message::BootstrapRequest(our_public_key, name_hash)) {
+            Ok(true) => EventSet::error() | EventSet::readable(),
+            Ok(false) => EventSet::error() | EventSet::writable(),
+            Err(error) => {
+                error!("Failed to write to socket: {:?}", error);
+                let _ = core.remove_context(token);
+                return;
+            }
+        };
+
         if let Err(error) = event_loop.register(&socket,
                                                 token,
-                                                EventSet::error() | EventSet::writable(),
+                                                event_set,
                                                 PollOpt::edge()) {
             error!("Failed to register socket: {:?}", error);
             let _ = core.remove_context(token);
--- a/src/bootstrap_states/establish_connection.rs
+++ b/src/bootstrap_states/establish_connection.rs
@@ -53,6 +53,11 @@ impl EstablishConnection {
                  parent_handle: Context,
                  event_tx: ::CrustEventSender) {
 
+        if contact_info.tcp_acceptors.is_empty() {
+            warn!("List of TCP acceptors in contact info is empty");
+            return;
+        }
+
         let context = core.get_new_context();
         let token = core.get_new_token();
         let _ = core.insert_context(token, context);
@@ -67,19 +72,9 @@ impl EstablishConnection {
             }
         };
 
-        let event_set = match socket.write(Message::BootstrapRequest(our_public_key, name_hash)) {
-            Ok(true) => EventSet::error() | EventSet::readable(),
-            Ok(false) => EventSet::error() | EventSet::writable(),
-            Err(error) => {
-                error!("Failed to write to socket: {:?}", error);
-                let _ = core.remove_context(token);
-                return;
-            }
-        };
-
         if let Err(error) = event_loop.register(&socket,
                                                 token,
-                                                event_set,
+                                                EventSet::error() | EventSet::writable(),
                                                 PollOpt::edge()) {
             error!("Failed to register socket: {:?}", error);
             let _ = core.remove_context(token);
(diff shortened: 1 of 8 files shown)


maidsafe-highfive:
r? @canndrew

ustulation:
Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/bootstrap_states/establish_connection.rs, line 77 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/707#-KIYBABCpGONrtjq93AI:-KIYBABCpGONrtjq93AJ:1332184098) ([raw file](https://github.com/maidsafe/crust/blob/a02224766666be4f35089f2c960dd20671aaa8a4/src/bootstrap_states/establish_connection.rs#L77)):*
> ```Rust
>         if let Err(error) = event_loop.register(&socket,
>                                                 token,
>                                                 EventSet::error() | EventSet::writable(),
> ```

We should cater for hup() too - hup() means a graceful shutdown and will events will not fire if that happens (only readable and hup get fired in hup i think - anyway it's better to use hup() too) {hup stands for hung-up btw for graceful connection closing}. Also we don't want to set the writable flag anywhere but in the write handler if more data is to be written. Sockets are usually writable and almost certainly so at the beginning. So this will just be a wasted cycle - instead the following paradigm suffices in all cases: register/reregister the socket with readable | error | hup at the start (in the constructor) always. If something is to be written immediately call the write method. In the write method depending on if more data is left to be written, reregister with either readable | error | hup OR with readable | writable | error | hup. So readable | error and hup are never taken out. Only writable is inserted or not inserted. It keeps management easier and also allows to fully leverage the sockets. Sockets are bidirectional so setting only readable or writable at a time is not ideal.

---

*[src/bootstrap_states/establish_connection.rs, line 184 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/707#-KIYCv9xn3lIFRr2_6h7:-KIYCv9xn3lIFRr2_6h8:265517543) ([raw file](https://github.com/maidsafe/crust/blob/a02224766666be4f35089f2c960dd20671aaa8a4/src/bootstrap_states/establish_connection.rs#L184)):*
> ```Rust
>     }
> 
>     fn set_readable(&mut self, core: &mut Core, event_loop: &mut EventLoop<Core>, token: Token) {
> ```

Due to earlier comment, this will be useful (called from) only in one place and as such not useful to have as a function call. So better to remove it

---

*[src/bootstrap_states/establish_connection.rs, line 188 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/707#-KIYDAOVSwSHvjk41wim:-KIYDAOVSwSHvjk41win:265517543) ([raw file](https://github.com/maidsafe/crust/blob/a02224766666be4f35089f2c960dd20671aaa8a4/src/bootstrap_states/establish_connection.rs#L188)):*
> ```Rust
>     }
> 
>     fn set_writable(&mut self, core: &mut Core, event_loop: &mut EventLoop<Core>, token: Token) {
> ```

Due to earlier comment, this will be useful (called from) only in one place and as such not useful to have as a function call. So better to remove it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/707)*

madadam:
Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/bootstrap_states/establish_connection.rs, line 77 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/707#-KIYBABCpGONrtjq93AI:-KIYEj5dp_01ryAuWWP_:-1615226911) ([raw file](https://github.com/maidsafe/crust/blob/a02224766666be4f35089f2c960dd20671aaa8a4/src/bootstrap_states/establish_connection.rs#L77)):*
> We should cater for hup() too - hup() means a graceful shutdown and will events will not fire if that happens (only readable and hup get fired in hup i think - anyway it's better to use hup() too) {hup stands for hung-up btw for graceful connection closing}. Also we don't want to set the writable flag anywhere but in the write handler if more data is to be written. Sockets are usually writable and almost certainly so at the beginning. So this will just be a wasted cycle - instead the following paradigm suffices in all cases: register/reregister the socket with readable | error | hup at the start (in the constructor) always. If something is to be written immediately call the write method. In the write method depending on if more data is left to be written, reregister with either readable | error | hup OR with readable | writable | error | hup. So readable | error and hup are never taken out. Only writable is inserted or not inserted. It keeps management easier and also allows to fully leverage the sockets. Sockets are bidirectional so setting only readable or writable at a time is not ideal.
</details>
Writing to the socket right after connecting it resulted in failure on osx before, so that's why I'm scheduling writable event here and writing only in the ready handler. Rest of your points is OK and I'll update the PR accordingly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/707)*

madadam:
Closing in favor of https://github.com/maidsafe/crust/pull/708

PR #711, opened 2016-05-25
feat/connection_listener: enables timeout when accepting connection

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/711)

### DIFF
--- a/src/bootstrap_states/bootstrap.rs
+++ b/src/bootstrap_states/bootstrap.rs
@@ -23,12 +23,13 @@ use std::collections::HashSet;
 use std::mem;
 use std::rc::Rc;
 
-use connection_listener::BOOTSTRAP_TIMEOUT_MS;
 use core::{Context, Core, State};
 use event::Event;
 use service::SharedConnectionMap;
 use static_contact_info::StaticContactInfo;
 use super::establish_connection::EstablishConnection;
+
+const BOOTSTRAP_TIMEOUT_MS: u64 = 60_000;
 
 pub struct Bootstrap {
     context: Context,
--- a/src/bootstrap_states/bootstrap.rs
+++ b/src/bootstrap_states/bootstrap.rs
@@ -23,13 +23,12 @@ use std::collections::HashSet;
 use std::mem;
 use std::rc::Rc;
 
+use connection_listener::BOOTSTRAP_TIMEOUT_MS;
 use core::{Context, Core, State};
 use event::Event;
 use service::SharedConnectionMap;
 use static_contact_info::StaticContactInfo;
 use super::establish_connection::EstablishConnection;
-
-const BOOTSTRAP_TIMEOUT_MS: u64 = 60_000;
 
 pub struct Bootstrap {
     context: Context,
--- a/src/bootstrap_states/bootstrap.rs
+++ b/src/bootstrap_states/bootstrap.rs
@@ -15,11 +15,12 @@
 // Please review the Licences for the specific language governing permissions and limitations
 // relating to use of the SAFE Network Software.
 
-use mio::EventLoop;
+use mio::{EventLoop, Timeout, Token};
 use sodiumoxide::crypto::box_::PublicKey;
 use std::any::Any;
 use std::cell::RefCell;
 use std::collections::HashSet;
+use std::mem;
 use std::rc::Rc;
 
 use core::{Context, Core, State};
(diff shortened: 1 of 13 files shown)


maidsafe-highfive:
r? @canndrew

ustulation:

---

*[src/bootstrap_states/bootstrap.rs, line 26 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/711#-KIcYQgugskvjEtWr0Jo:-KIcYQgugskvjEtWr0Jp:2056887448) ([raw file](https://github.com/maidsafe/crust/blob/0be348b397fa053079323139760092d2d3e0dce4/src/bootstrap_states/bootstrap.rs#L26)):*
> ```Rust
> use std::rc::Rc;
> 
> use connection_listener::BOOTSTRAP_TIMEOUT_MS;
> ```

having bootstrap module depend on connection listener module doesn't make sense. It is better to keep this in service.rs if both depend on this constant

---

*[src/core/state.rs, line 46 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/711#-KIcYk0NKQtRhZ-s2EFD:-KIcYk0NKQtRhZ-s2EFE:1311172379) ([raw file](https://github.com/maidsafe/crust/blob/0be348b397fa053079323139760092d2d3e0dce4/src/core/state.rs#L46)):*
> ```Rust
>                _event_loop: &mut EventLoop<Core>,
>                _token: Token) {}
> ```

Na it's better to keep it as what it was previously. Putting Token everywhere will cause us to re-do it in all places should there be a change. state.rs shouldn't care what it is - it depends on core so whatever core mentions about the type should be good enough. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/711)*

ustulation:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---

*[src/bootstrap_states/bootstrap.rs, line 26 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/711#-KIcYQgugskvjEtWr0Jo:-KIc_NKt8V4dnEvcpMGm:44682308) ([raw file](https://github.com/maidsafe/crust/blob/0be348b397fa053079323139760092d2d3e0dce4/src/bootstrap_states/bootstrap.rs#L26)):*
> having bootstrap module depend on connection listener module doesn't make sense. It is better to keep this in service.rs if both depend on this constant
</details>
rather these two infact should be independent of each other - they have no correlation - bootstrap time out means how long should bootstrap be attempted for before giving up. Timeout in connection listener means how long should it wait for the required messages before giving up. They have nothing in common even if they share the same value - pls separate the constants

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/711)*

maqi:
*[src/bootstrap_states/bootstrap.rs, line 26 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/711#-KIcYQgugskvjEtWr0Jo:-KIcaf7Lu50bY_CFO-58:940540440) ([raw file](https://github.com/maidsafe/crust/blob/0be348b397fa053079323139760092d2d3e0dce4/src/bootstrap_states/bootstrap.rs#L26)):*
> rather these two infact should be independent of each other - they have no correlation - bootstrap time out means how long should bootstrap be attempted for before giving up. Timeout in connection listener means how long should it wait for the required messages before giving up. They have nothing in common even if they share the same value - pls separate the constants
</details>
but the Timeout in connection listener is for per connection. there is no point if the request has already dropped but the listener is still waiting for that, or the listener is closed but the request is still waiting.
unless you are talking about the timeout for the listener as the whole.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/711)*

maqi:
*[src/core/state.rs, line 46 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/711#-KIcYk0NKQtRhZ-s2EFD:-KIcbBBh3KL-tEMyIh4v:-1529267473) ([raw file](https://github.com/maidsafe/crust/blob/0be348b397fa053079323139760092d2d3e0dce4/src/core/state.rs#L46)):*
> Na it's better to keep it as what it was previously. Putting Token everywhere will cause us to re-do it in all places should there be a change. state.rs shouldn't care what it is - it depends on core so whatever core mentions about the type should be good enough. 
</details>
i removed it in my previous commits, but after rebase, seems the bootstrap timeout wants it back.
do you mean it is now decided this token to be removed anyway?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/711)*

PR #724, opened 2016-05-26
Direct connect

madadam:
---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/724)

### DIFF
--- a/src/active_connection.rs
+++ b/src/active_connection.rs
@@ -19,10 +19,10 @@ use mio::{EventLoop, EventSet, PollOpt, Timeout, TimerError, Token};
 use std::any::Any;
 
 use core::{Core, Context, State};
+use connect::SharedConnectionMap;
 use event::Event;
 use message::Message;
 use peer_id::PeerId;
-use service::SharedConnectionMap;
 use socket::Socket;
 use std::rc::Rc;
 use std::cell::RefCell;
--- a/src/active_connection.rs
+++ b/src/active_connection.rs
@@ -179,10 +179,6 @@ impl State for ActiveConnection {
     fn terminate(&mut self, core: &mut Core, event_loop: &mut EventLoop<Core>) {
         self.heartbeat.terminate(core, event_loop);
 
-        if let Err(error) = self.socket.shutdown() {
-            debug!("Failed to shutdown socket: {:?}", error);
-        }
-
         if let Err(error) = event_loop.deregister(&self.socket) {
             debug!("Failed to deregister socket: {:?}", error);
         }
--- a/src/bootstrap/mod.rs
+++ b/src/bootstrap/mod.rs
@@ -18,28 +18,28 @@
 mod cache;
 mod try_peer;
 
-use std::net;
-use std::mem;
+use mio::{EventLoop, Timeout, Token};
+use socket_addr;
+use sodiumoxide::crypto::box_::PublicKey;
 use std::any::Any;
 use std::cell::RefCell;
+use std::collections::HashSet;
+use std::mem;
+use std::net;
 use std::rc::{Rc, Weak};
 use std::sync::mpsc::{self, Receiver};
 
-use socket_addr;
-use event::Event;
-use socket::Socket;
-use peer_id::PeerId;
-use error::CrustError;
-use self::cache::Cache;
-use config_handler::Config;
-use self::try_peer::TryPeer;
-use std::collections::HashSet;
-use core::{Context, Core, State};
-use service::SharedConnectionMap;
-use mio::{EventLoop, Timeout, Token};
-use service_discovery::ServiceDiscovery;
 use active_connection::ActiveConnection;
-use sodiumoxide::crypto::box_::PublicKey;
+use config_handler::Config;
+use connect::SharedConnectionMap;
+use core::{Context, Core, State};
+use error::CrustError;
+use event::Event;
+use peer_id::PeerId;
+use socket::Socket;
+use self::cache::Cache;
+use self::try_peer::TryPeer;
+use service_discovery::ServiceDiscovery;
 use static_contact_info::StaticContactInfo;
 
 const BOOTSTRAP_TIMEOUT_MS: u64 = 10000;
--- a/src/connect/connection_candidate.rs
+++ b/src/connect/connection_candidate.rs
@@ -1,0 +1,186 @@
+// Copyright 2016 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under (1) the MaidSafe.net Commercial License,
+// version 1.0 or later, or (2) The General Public License (GPL), version 3, depending on which
+// licence you accepted on initial access to the Software (the "Licences").
+//
+// By contributing code to the SAFE Network Software, or to this project generally, you agree to be
+// bound by the terms of the MaidSafe Contributor Agreement, version 1.0.  This, along with the
+// Licenses can be found in the root directory of this project at LICENSE, COPYING and CONTRIBUTOR.
+//
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.
+//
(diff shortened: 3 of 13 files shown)


maidsafe-highfive:
r? @vinipsmaker

ustulation:
Review status: 0 of 13 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/connection_listener/exchange_msg.rs, line 99 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/724#-KJ6IOE8XtxUK5Aje_Qt:-KJ6IOE8XtxUK5Aje_Qu:-984245984) ([raw file](https://github.com/maidsafe/crust/blob/65dabde1b267d64c9d106621de028c781871ca1c/src/connection_listener/exchange_msg.rs#L99)):*
> ```Rust
>     }
> 
>     fn writable(&mut self, core: &mut Core, event_loop: &mut EventLoop<Core>) {
> ```

this function has a one liner body and is called from only one place in the file. Can we need this indirection?

---

*[src/connection_listener/exchange_msg.rs, line 197 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/724#-KJ6IhkXyHEpiQC4weKi:-KJ6IhkXyHEpiQC4weKj:775237930) ([raw file](https://github.com/maidsafe/crust/blob/65dabde1b267d64c9d106621de028c781871ca1c/src/connection_listener/exchange_msg.rs#L197)):*
> ```Rust
>                                            self.event_tx.clone());
>             }
>             NextState::None => unreachable!(),
> ```

Why do we even have this option ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/724)*

madadam:
*[src/connection_listener/exchange_msg.rs, line 197 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/724#-KJ6IhkXyHEpiQC4weKi:-KJ6JO7Qo9iOlUUKJdxg:-542795326) ([raw file](https://github.com/maidsafe/crust/blob/65dabde1b267d64c9d106621de028c781871ca1c/src/connection_listener/exchange_msg.rs#L197)):*
> Why do we even have this option ?
</details>
It's the initial value of the field. The other variants contain the peer id, which I don't know from the beginning. Alternative would be to change the field type to Option<NextState> and assign None to it, but then we would still have to handle the `None` case somehow.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/724)*

madadam:
*[src/connection_listener/exchange_msg.rs, line 99 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/724#-KJ6IOE8XtxUK5Aje_Qt:-KJ6JrlI9ngBoakyVWxe:597277060) ([raw file](https://github.com/maidsafe/crust/blob/65dabde1b267d64c9d106621de028c781871ca1c/src/connection_listener/exchange_msg.rs#L99)):*
> this function has a one liner body and is called from only one place in the file. Can we need this indirection?
</details>
I'll remove it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/724)*

ustulation:
*[src/connection_listener/exchange_msg.rs, line 197 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/724#-KJ6IhkXyHEpiQC4weKi:-KJ6MCJW_Whby-FZZ3RV:2072271917) ([raw file](https://github.com/maidsafe/crust/blob/65dabde1b267d64c9d106621de028c781871ca1c/src/connection_listener/exchange_msg.rs#L197)):*
<details><summary><i>Previously, madadam (Adam Cigánek) wrote…</i></summary>
> It's the initial value of the field. The other variants contain the peer id, which I don't know from the beginning. Alternative would be to change the field type to Option<NextState> and assign None to it, but then we would still have to handle the `None` case somehow.
</details>
Not necessarily - it's pretty easy - let the default state be transitioning to ActiveConnection (that's what the bootstrap does) if successful , for the case of `Connect` you can put the other variant. OR the other way around - it doesn't matter. It will save unreachable codes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/724)*

madadam:
*[src/connection_listener/exchange_msg.rs, line 197 \[r3\]](https://reviewable.io:443/reviews/maidsafe/crust/724#-KJ6IhkXyHEpiQC4weKi:-KJ6NvfyJSz1Y8CTvHdv:1946656770) ([raw file](https://github.com/maidsafe/crust/blob/65dabde1b267d64c9d106621de028c781871ca1c/src/connection_listener/exchange_msg.rs#L197)):*
> Not necessarily - it's pretty easy - let the default state be transitioning to ActiveConnection (that's what the bootstrap does) if successful , for the case of `Connect` you can put the other variant. OR the other way around - it doesn't matter. It will save unreachable codes.
</details>
We still have to store the peer_id somewhere.  There was a separate `Option<PeerId>` field for it previously, but I feel storing the peer_id in the same variable as the next state makes sense, because the peer_id becomes available precisely when we know what the next state should be so it's not possible to make a mistake of setting one, but not the other.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/724)*

ustulation:
*[src/connect/connection_candidate.rs, line 55 \[r4\]](https://reviewable.io:443/reviews/maidsafe/crust/724#-KJ6QBjk3NsD7IpnDcZe:-KJ6QBjk3NsD7IpnDcZf:965655342) ([raw file](https://github.com/maidsafe/crust/blob/2a8fb52df5bd1933941e8298068a7b67b1880b5c/src/connect/connection_candidate.rs#L55)):*
> ```Rust
> 
>         let result = if our_id > their_id {
>             event_loop.reregister(&socket, token, event_set | EventSet::writable(), PollOpt::edge())
> ```

Do we need this? Cant we go and wirte directly - we have come to this stage _after_ connection is successful afterall

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/724)*

ustulation:
*[src/connect/connection_candidate.rs, line 50 \[r4\]](https://reviewable.io:443/reviews/maidsafe/crust/724#-KJ6QQ6nD1qfiOIlhCUJ:-KJ6QQ6nD1qfiOIlhCUK:1915740502) ([raw file](https://github.com/maidsafe/crust/blob/2a8fb52df5bd1933941e8298068a7b67b1880b5c/src/connect/connection_candidate.rs#L50)):*
> ```Rust
>                  event_tx: ::CrustEventSender)
>     {
>         let event_set = EventSet::error() |
> ```

If the comment for a few lines below holds, this is unnecessary

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/724)*

ustulation:
*[src/connect/connection_candidate.rs, line 60 \[r4\]](https://reviewable.io:443/reviews/maidsafe/crust/724#-KJ6QcneUZEIfjegRAEs:-KJ6QcneUZEIfjegRAEt:-1299018471) ([raw file](https://github.com/maidsafe/crust/blob/2a8fb52df5bd1933941e8298068a7b67b1880b5c/src/connect/connection_candidate.rs#L60)):*
> ```Rust
>         };
> 
>         if let Err(error) = result {
> ```

this is not required if the previous comments hold

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/724)*

ustulation:
*[src/connect/connection_candidate.rs, line 100 \[r4\]](https://reviewable.io:443/reviews/maidsafe/crust/724#-KJ6QqxR4ZW7YFZC_c22:-KJ6QqxR4ZW7YFZC_c23:679200989) ([raw file](https://github.com/maidsafe/crust/blob/2a8fb52df5bd1933941e8298068a7b67b1880b5c/src/connect/connection_candidate.rs#L100)):*
> ```Rust
>     fn writable(&mut self, core: &mut Core, event_loop: &mut EventLoop<Core>) {
>         if self.connection_exists() {
>             self.terminate(core, event_loop);
> ```

this comment is trivial so feel free to ignore it if you don't agree - is it not better to have it succinctly as `return self.terminate(core, event_loop)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/724)*

PR #725, opened 2016-05-26
refactor/connections: better code

ustulation:
1) using write_2 instead of write in socket.rs - this solves more issues and gets better code in places where it is used 2) fix timer cancellation on promoting to active connection from exchange_msg 3) new file exchange_msg is much cleaner and smaller than the original counterpart accept_connections

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/725)


maidsafe-highfive:
r? @canndrew

PR #729, opened 2016-05-27
test/connection-listener: extensive test cases on all platforms & osx…

ustulation:
… behaviour

1) Extensively test all behaviours of connection listener including expected failure cases
2) When A and B are connected and B closes connection gracefully, A's subsequent write to B does not fail on OSX. It fails after some attempts. Read however always returns EOF on all platforms. So use that in test cases to properly validate exchange msg timeouts.

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/729)


maidsafe-highfive:
r? @canndrew

PR #730, opened 2016-05-27
fix/test: handle OS dependent behaviour

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/730)


maidsafe-highfive:
r? @vinipsmaker

PR #731, opened 2016-05-27
fix OS dependent issue

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/731)


maidsafe-highfive:
r? @vinipsmaker

PR #735, opened 2016-05-31
Use Socket::write_2 (renamed to write) everywhere

madadam:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/735)

### DIFF
--- a/src/active_connection.rs
+++ b/src/active_connection.rs
@@ -45,7 +45,6 @@ pub struct ActiveConnection {
     peer_id: PeerId,
     event_tx: ::CrustEventSender,
     heartbeat: Heartbeat,
-    event_set: EventSet,
 }
 
 impl ActiveConnection {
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @canndrew

ustulation:
*[src/socket.rs, line 176 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/735#-KJ625OuLYLPEfj10_0U:-KJ625Ov6I13cO8VcRxw:-700609590) ([raw file](https://github.com/maidsafe/crust/blob/78759ac2101097c22db16e29dfe501ca39f779b6/src/socket.rs#L176)):*
> ```Rust
>                         } else {
>                             EventSet::writable()
>                         };
> ```

I don't think this is more readable or prettier than the previous one. There is no significant improvement (in my opinion in particular there's no improvement at all) and uses 3 extra lines of code. Should we revert this ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/735)*

PR #740, opened 2016-06-01
test/service: bring back the multiple services test

afck:
This is ignored for now. Once it works, it should probably be turned
into an integration test.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/740)


maidsafe-highfive:
r? @vinipsmaker

PR #741, opened 2016-06-01
Bug fixes and also now multiple services test runs with 15 nodes

ustulation:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/741)


maidsafe-highfive:
r? @vinipsmaker

PR #746, opened 2016-06-03
fix/event: do not raise spurious NewPeer(Err(_), _)

ustulation:
1) Extend connection map to account for currently handshaking connections
2) Check if we have exhausted all the given endpoints for the peer AND after that check if none of the connections are in handshaking stage or have become an active connection, before raising NewPeer(Err)

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/746)

### DIFF
--- a/src/active_connection.rs
+++ b/src/active_connection.rs
@@ -18,7 +18,7 @@
 use std::any::Any;
 
 use core::{Core, Context, State};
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @vinipsmaker

PR #747, opened 2016-06-03
fix/event: fix new-peer-error edge case - important commit

ustulation:
This is an interesting commit. So basically each time routing ci test was run with large number of nodes (e.g. 50) we use to see some occasional NewPeer(Err) which should be virtually impossible to get after the previous fix at-least on the same computer (localhost). It turns out the the backlog value of TCP Listener was set to 1. What this does is silently ignore incoming TCP 3-way handshake if there is one pending handshake already in the kernel's backlog. So although our code is resilient to spurious NewPeer(Err) the kernel never informs about an incoming SYN and thus we get a few NewPeer(Err) for no apparant reason. This commit puts that to 100 currently. Now with this the system is slow so there is a call to accept in a loop till we get a spurious accept at which we go back to event loop and this compensates for the slowness (balances out). 100 is chosen because https://www.ibm.com/support/knowledgecenter/SSFKSJ_9.0.0/com.ibm.mq.con.doc/q016110_.htm

---

This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/747)

### DIFF
--- a/src/connection_listener/mod.rs
+++ b/src/connection_listener/mod.rs
@@ -37,6 +37,8 @@ use nat::mapping_context::MappingContext;
 use socket::Socket;
 use static_contact_info::StaticContactInfo;
 
+const LISTENER_BACKLOG: i32 = 100;
+
 pub struct ConnectionListener {
     cm: ConnectionMap,
     context: Context,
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @vinipsmaker

PR #754, opened 2016-06-09
feat/priority: implement message prioritization

ustulation:
All messages to be sent out now need to specify priority. Crust will do it best to ensure the higher priority (0 being the highest) messages are sent before any lower priority ones even if lower priority were the ones to come into the queue first. Priority ranges between 0 (highest) and 255 (lowest).

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/754)

### DIFF
--- a/src/active_connection.rs
+++ b/src/active_connection.rs
@@ -18,7 +18,7 @@
 use std::any::Any;
 use std::time::Duration;
 
-use core::{Core, Context, State};
+use core::{Core, Context, State, Priority};
 use service::{ConnectionId, ConnectionMap};
 use event::Event;
 use message::Message;
@@ -128,7 +128,10 @@ impl ActiveConnection {
         }
     }
 
-    fn write(&mut self, core: &mut Core, event_loop: &mut EventLoop<Core>, msg: Option<Message>) {
+    fn write(&mut self,
+             core: &mut Core,
+             event_loop: &mut EventLoop<Core>,
+             msg: Option<(Message, Priority)>) {
         if let Err(error) = self.socket.write(event_loop, self.token, msg) {
             debug!("{:?} - Failed to write socket: {:?}", self.our_id, error);
             self.terminate(core, event_loop);
@@ -168,8 +171,12 @@ impl State for ActiveConnection {
         }
     }
 
-    fn write(&mut self, core: &mut Core, event_loop: &mut EventLoop<Core>, data: Vec<u8>) {
-        self.write(core, event_loop, Some(Message::Data(data)));
+    fn write(&mut self,
+             core: &mut Core,
+             event_loop: &mut EventLoop<Core>,
+             data: Vec<u8>,
+             priority: Priority) {
+        self.write(core, event_loop, Some((Message::Data(data), priority)));
         self.reset_send_heartbeat(core, event_loop);
     }
 
@@ -204,7 +211,7 @@ impl State for ActiveConnection {
     fn timeout(&mut self, core: &mut Core, event_loop: &mut EventLoop<Core>, token: Token) {
         match self.heartbeat.timeout(event_loop, token) {
             HeartbeatAction::None => (),
-            HeartbeatAction::Send => self.write(core, event_loop, Some(Message::Heartbeat)),
+            HeartbeatAction::Send => self.write(core, event_loop, Some((Message::Heartbeat, 0))),
             HeartbeatAction::Terminate => {
                 // TODO Disabling heartbeat for now to make testing easier
                 // debug!("Dropping connection to {:?} due to peer inactivity",
@@ -233,11 +240,13 @@ struct Heartbeat {
 impl Heartbeat {
     fn new(core: &mut Core, event_loop: &mut EventLoop<Core>, context: Context) -> ::Res<Self> {
         let recv_token = core.get_new_token();
-        let recv_timeout = try!(event_loop.timeout(recv_token, Duration::from_millis(INACTIVITY_TIMEOUT_MS)));
+        let recv_timeout =
+            try!(event_loop.timeout(recv_token, Duration::from_millis(INACTIVITY_TIMEOUT_MS)));
         let _ = core.insert_context(recv_token, context);
 
         let send_token = core.get_new_token();
-        let send_timeout = try!(event_loop.timeout(send_token, Duration::from_millis(HEARTBEAT_PERIOD_MS)));
+        let send_timeout =
+            try!(event_loop.timeout(send_token, Duration::from_millis(HEARTBEAT_PERIOD_MS)));
         let _ = core.insert_context(send_token, context);
 
         Ok(Heartbeat {
@@ -253,8 +262,8 @@ impl Heartbeat {
         //     return HeartbeatAction::Terminate;
         // }
         if token == self.send_token {
-            return if let Err(error) =
-                          event_loop.timeout(self.send_token, Duration::from_millis(HEARTBEAT_PERIOD_MS)) {
+            let timeout = Duration::from_millis(HEARTBEAT_PERIOD_MS);
+            return if let Err(error) = event_loop.timeout(self.send_token, timeout) {
                 warn!("Failed to reschedule heartbeat send timer: {:?}", error);
                 HeartbeatAction::Terminate
             } else {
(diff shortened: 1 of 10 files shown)


maidsafe-highfive:
r? @canndrew

afck:
Review status: 0 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/socket.rs, line 150 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/754#-KJovv6EByeZw0rKGDZU:-KJovv6FD5RHLYRbNZnz:-686478470) ([raw file](https://github.com/maidsafe/crust/blob/980e47d98f6c7bee2d77d0bba4aa023761873696/src/socket.rs#L150)):*
> ```Rust
>             try!(data.write_u32::<LittleEndian>(len as u32));
> 
>             let entry = self.write_queue.entry(priority).or_insert(VecDeque::with_capacity(10));
> ```

This will create a `VecDeque` whether the entry exists or not. It's better to use `or_insert_with` and `VecDeque::new` or `|| VecDeque::with_capacity(10)`.

---

*[src/core/state.rs, line 24 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/754#-KJowZKOP9WEmHKzzFW_:-KJowZKPqjQx3YXdmBms:523867920) ([raw file](https://github.com/maidsafe/crust/blob/980e47d98f6c7bee2d77d0bba4aa023761873696/src/core/state.rs#L24)):*
> ```Rust
> use mio::{EventLoop, EventSet, Handler, Token};
> 
> /// Priority of a message to be sent by Crust. Priority 0 being the highest and very unlikely to be
> ```

It will probably make sense to actually _never_ drop priority 0 (that's what we currently do in master).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/754)*

ustulation:
*[src/socket.rs, line 150 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/754#-KJovv6EByeZw0rKGDZU:-KJp-sYLDp_W2r_UZL9M:-440690615) ([raw file](https://github.com/maidsafe/crust/blob/980e47d98f6c7bee2d77d0bba4aa023761873696/src/socket.rs#L150)):*
> This will create a VecDeque whether the entry exists or not
But that is what i want - if the entry does not exist create a default one so that i can do a push back in the next line below. Sorry i didn't get this point could you explain more ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/754)*

ustulation:
*[src/socket.rs, line 150 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/754#-KJovv6EByeZw0rKGDZU:-KJp0N5w1Uw7B1k6Dh7D:-1299724977) ([raw file](https://github.com/maidsafe/crust/blob/980e47d98f6c7bee2d77d0bba4aa023761873696/src/socket.rs#L150)):*
> This will create a VecDeque whether the entry exists or not

But that is what i want - if the entry does not exist create a default one so that i can do a push back in the next line below. Sorry i didn't get this point could you explain more ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/754)*

afck:
Review status: 0 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/socket.rs, line 150 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/754#-KJovv6EByeZw0rKGDZU:-KJp0UibJHztG-UUmEps:55218022) ([raw file](https://github.com/maidsafe/crust/blob/980e47d98f6c7bee2d77d0bba4aa023761873696/src/socket.rs#L150)):*
> > This will create a VecDeque whether the entry exists or not
> 
> But that is what i want - if the entry does not exist create a default one so that i can do a push back in the next line below. Sorry i didn't get this point could you explain more ?
> 
</details>
If the entry doesn't exist it's fine. But if it _does_ exist, this will allocate a new `VecDeque` anyway and pass it ot the `or_insert` method, just to have it dropped there again.
`or_insert_with(VecDeque::new)` would only call the `VecDeque::new` constructor if it actually needs it, i. e. if the entry actually doesn't exist, avoiding the unnecessary allocation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/754)*

ustulation:
*[src/socket.rs, line 150 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/754#-KJovv6EByeZw0rKGDZU:-KJp19FSL2M74rHmsutQ:-166095123) ([raw file](https://github.com/maidsafe/crust/blob/980e47d98f6c7bee2d77d0bba4aa023761873696/src/socket.rs#L150)):*
<details><summary><i>Previously, afck (Andreas Fackler) wrote…</i></summary>
> If the entry doesn't exist it's fine. But if it _does_ exist, this will allocate a new `VecDeque` anyway and pass it ot the `or_insert` method, just to have it dropped there again.
> `or_insert_with(VecDeque::new)` would only call the `VecDeque::new` constructor if it actually needs it, i. e. if the entry actually doesn't exist, avoiding the unnecessary allocation.
</details>
I dont think that is how it works (either that or my head isnt't functioning correctly)
https://doc.rust-lang.org/nightly/std/collections/btree_map/enum.Entry.html#method.or_insert
So from that - if the entry does not exist it will use the default, if it does then it will return the existing. That is what _default_ means i hope in that function parameter

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/754)*

afck:
Review status: 0 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/socket.rs, line 150 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/754#-KJovv6EByeZw0rKGDZU:-KJp252m2go3zOUePkaG:505225277) ([raw file](https://github.com/maidsafe/crust/blob/980e47d98f6c7bee2d77d0bba4aa023761873696/src/socket.rs#L150)):*
> I dont think that is how it works (either that or my head isnt't functioning correctly)
> https://doc.rust-lang.org/nightly/std/collections/btree_map/enum.Entry.html#method.or_insert
> So from that - if the entry does not exist it will use the default, if it does then it will return the existing. That is what _default_ means i hope in that function parameter
</details>
That's what the `or_insert` method does, yes. However, it is _called_ in either case, and as an argument it takes the newly (and often unnecessarily) allocated `VecDeque`. It then either inserts it or drops it, which is a waste.
The `or_insert_with` method takes an `FnOnce` instead and only calls that if the entry is empty. If that `FnOnce` is where the new `VecDeque` is being allocated, the allocation actually is avoided if the entry is empty.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/754)*

ustulation:
Review status: 0 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/socket.rs, line 150 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/754#-KJovv6EByeZw0rKGDZU:-KJp5UCn1r6DXOOUfnaR:2125488977) ([raw file](https://github.com/maidsafe/crust/blob/980e47d98f6c7bee2d77d0bba4aa023761873696/src/socket.rs#L150)):*
<details><summary><i>Previously, afck (Andreas Fackler) wrote…</i></summary>
> That's what the `or_insert` method does, yes. However, it is _called_ in either case, and as an argument it takes the newly (and often unnecessarily) allocated `VecDeque`. It then either inserts it or drops it, which is a waste.
> The `or_insert_with` method takes an `FnOnce` instead and only calls that if the entry is empty. If that `FnOnce` is where the new `VecDeque` is being allocated, the allocation actually is avoided if the entry is empty.
</details>
Ah right i see what you mean - i've used a lamda instead now

---

*[src/core/state.rs, line 24 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/754#-KJowZKOP9WEmHKzzFW_:-KJp0jyMa-yOA8gXC1OW:677588487) ([raw file](https://github.com/maidsafe/crust/blob/980e47d98f6c7bee2d77d0bba4aa023761873696/src/core/state.rs#L24)):*
<details><summary><i>Previously, afck (Andreas Fackler) wrote…</i></summary>
> It will probably make sense to actually _never_ drop priority 0 (that's what we currently do in master).
</details>
done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/754)*

afck:
Reviewed 8 of 10 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/754)*

PR #760, opened 2016-06-13
feat/core: also drop messages with higher priorities if queue expires

afck:
If e.g. a message with priority 2 expires, drop all messages with
priority >= 2.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/760)

### DIFF
--- a/src/common/socket.rs
+++ b/src/common/socket.rs
@@ -131,24 +131,25 @@ impl Socket {
                                el: &mut EventLoop<Core>,
                                token: Token,
                                msg: Option<(T, Priority)>)
-                               -> Result<bool> {
-        let mut expired_keys = Vec::new();
-        let _ = self.write_queue
+                               -> ::Res<bool> {
+        let expired_keys: Vec<u8> = self.write_queue
             .iter()
-            .all(|(&priority, ref queue)| {
-                if priority != 0 && // Don't drop messages with priority 0.
-                           queue.front().map_or(true, |&(ref timestamp, _)| {
-                               timestamp.elapsed().as_secs() > MAX_MSG_AGE_SECS
-                           }) {
-                    debug!("Insufficient bandwidth. Dropping {} messages with priority {}.",
-                           queue.len(),
-                           priority);
-                    expired_keys.push(priority);
-                }
-                true
-            });
-        for it in expired_keys.iter() {
-            let _ = self.write_queue.remove(&it);
+            .skip_while(|&(&priority, ref queue)| {
+                priority == 0 || // Don't drop messages with priority 0.
+                queue.front().map_or(false, |&(ref timestamp, _)| {
+                    timestamp.elapsed().as_secs() <= MAX_MSG_AGE_SECS
+                })
+            })
+            .map(|(&priority, _)| priority)
+            .collect();
+        let dropped_msgs: usize = expired_keys.iter()
+            .filter_map(|priority| self.write_queue.remove(priority))
+            .map(|queue| queue.len())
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @canndrew

ustulation:
hey @afck can you pls pull master and add this change again

ustulation:
*[src/common/socket.rs, line 139 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/760#-KKD5PvG-md1OxUAHE1A:-KKD5PvG-md1OxUAHE1B:-951134602) ([raw file](https://github.com/maidsafe/crust/blob/522f478d78a29aae32a3dcb95b4730e0518f59a2/src/common/socket.rs#L139)):*
> ```Rust
>             .skip_while(|&(&priority, ref queue)| {
>                 priority == 0 || // Don't drop messages with priority 0.
>                 queue.front().map_or(true, |&(ref timestamp, _)| {
> ```

shouldn't this be `map_or(false` ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/760)*

afck:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/common/socket.rs, line 139 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/760#-KKD5PvG-md1OxUAHE1A:-KKD5jQyNlwmDOOy_vCN:1665491985) ([raw file](https://github.com/maidsafe/crust/blob/522f478d78a29aae32a3dcb95b4730e0518f59a2/src/common/socket.rs#L139)):*
> shouldn't this be `map_or(false` ?
</details>
I was wondering whether it should even be `expect` - there should never be an empty entry in the `write_queue`, right?
But to be on the safe side, we can also go with `false`: That would drop the empty queue and all following ones. Not sure what's the best approach here?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/760)*

ustulation:
*[src/common/socket.rs, line 139 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/760#-KKD5PvG-md1OxUAHE1A:-KKDCFkUXcYgXhb9gEsY:1660550346) ([raw file](https://github.com/maidsafe/crust/blob/522f478d78a29aae32a3dcb95b4730e0518f59a2/src/common/socket.rs#L139)):*
<details><summary><i>Previously, afck (Andreas Fackler) wrote…</i></summary>
> I was wondering whether it should even be `expect` - there should never be an empty entry in the `write_queue`, right?
> But to be on the safe side, we can also go with `false`: That would drop the empty queue and all following ones. Not sure what's the best approach here?
</details>
ya lets expect is fine too - your call

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/760)*

afck:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/common/socket.rs, line 139 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/760#-KKD5PvG-md1OxUAHE1A:-KKDGaBsBcQCuymvmlQr:-499146009) ([raw file](https://github.com/maidsafe/crust/blob/522f478d78a29aae32a3dcb95b4730e0518f59a2/src/common/socket.rs#L139)):*
> ya lets expect is fine too - your call
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/760)*

PR #764, opened 2016-06-14
Crust update to mio

Viv-Rajkumar:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/764)


maidsafe-highfive:
r? @vinipsmaker

PR #781, opened 2016-06-21
feat/service-disc: flip service discovery

ustulation:
Fail it only for non-tests. For `#[cfg(test)]` there is no change.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/781)


ustulation:
Reviewed 1 of 1 files at r1, 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/781)*

PR #784, opened 2016-06-24
feat/socket: never drop priority 1 messages

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/784)

### DIFF
--- a/src/common/get_if_addrs.rs
+++ b/src/common/get_if_addrs.rs
@@ -40,7 +40,7 @@ pub struct Ifv6Addr {
     /// The netmask of the interface.
     pub netmask: Ipv6Addr,
     /// The broadcast address of the interface.
-    pub broadcast: Option<Ipv6Addr>
+    pub broadcast: Option<Ipv6Addr>,
 }
 
 impl Interface {
@@ -90,11 +90,11 @@ impl Ifv6Addr {
 #[cfg(not(windows))]
 mod getifaddrs_posix {
     use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
-    use std::{mem, io};
+    use std::{io, mem};
     use std::ffi::CStr;
 
     use c_linked_list::CLinkedListMut;
-    use common::get_if_addrs::{Interface, IfAddr, Ifv4Addr, Ifv6Addr};
+    use common::get_if_addrs::{IfAddr, Ifv4Addr, Ifv6Addr, Interface};
     use libc::{AF_INET, AF_INET6};
     use libc::getifaddrs as posix_getifaddrs;
     use libc::freeifaddrs as posix_freeifaddrs;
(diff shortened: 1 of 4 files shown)


ustulation:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/common/socket.rs, line 34 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/784#-KL0viiYqLIOGtEUkCDU:-KL0viiZ9xrNYVv0szNV:-1048085980) ([raw file](https://github.com/maidsafe/crust/blob/101e30642e70eadbfaeaafad4df613d594921baa/src/common/socket.rs#L34)):*
> ```Rust
> const MAX_MSG_AGE_SECS: u64 = 60;
> // Minimum priority for droppable messages. Messages with lower values will never be dropped.
> const MSG_DROP_PRIORITY: u8 = 2;
> ```

The priority is a public field in `send` function. So the user should be aware of what it means. So i think this should be a public constant in `lib.rs` otherwise we are keeping the users in dark. Also might be worthwhile updating the docs for `common::Priority` now to just say it indicates a message priority instead of more details like `Do not drop 0` etc it says right now

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/784)*

ustulation:

---

*[src/common/mod.rs, line 44 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/784#-KL11RyeCII9dhxY3yEg:-KL11RyeCII9dhxY3yEh:-315509439) ([raw file](https://github.com/maidsafe/crust/blob/d4f522a22742f1f1fd014672577de0c68a8e067b/src/common/mod.rs#L44)):*
> ```Rust
> mod state;
> 
> pub use self::socket::MSG_DROP_PRIORITY;
> ```

This should be move up and combined with `pub use self::socket::Socket` already present above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/784)*

afck:

---

*[src/common/mod.rs, line 44 \[r2\]](https://reviewable.io:443/reviews/maidsafe/crust/784#-KL11RyeCII9dhxY3yEg:-KL12TE57xHptfuVD_bI:1962150743) ([raw file](https://github.com/maidsafe/crust/blob/d4f522a22742f1f1fd014672577de0c68a8e067b/src/common/mod.rs#L44)):*
> This should be move up and combined with `pub use self::socket::Socket` already present above.
</details>
Actually, it should probably be defined here, together with `Priority`.

---

*[src/common/socket.rs, line 34 \[r1\]](https://reviewable.io:443/reviews/maidsafe/crust/784#-KL0viiYqLIOGtEUkCDU:-KL10kZ8Nd1K_O6qlInp:-499146009) ([raw file](https://github.com/maidsafe/crust/blob/101e30642e70eadbfaeaafad4df613d594921baa/src/common/socket.rs#L34)):*
> The priority is a public field in `send` function. So the user should be aware of what it means. So i think this should be a public constant in `lib.rs` otherwise we are keeping the users in dark. Also might be worthwhile updating the docs for `common::Priority` now to just say it indicates a message priority instead of more details like `Do not drop 0` etc it says right now
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/784)*

ustulation:
Reviewed 2 of 4 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/784)*

PR #785, opened 2016-06-24
feat/service-disc: revert flipping of service disc

ustulation:
Revert the condition which disallowed crust to bootstrap if it finds other instances running on LAN. Instead expose functionality for upper layers to determine the usage.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/785)


ustulation:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/crust/785)*

PR #788, opened 2016-06-29
Use the unwrap crate

canndrew:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/788)


ustulation:
Addressed in https://github.com/maidsafe/crust/pull/790

PR #790, opened 2016-06-29
Remove use of thread! macro

canndrew:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/crust/790)


ustulation:
https://github.com/maidsafe/crust/pull/791

PR #806, opened 2016-11-21
Fix build for rust 1.13 compiler

dhardy:


Viv-Rajkumar:
handled in a different way as part of #807

PR #810, opened 2017-01-18
Clippy

vinipsmaker:


Cargo.toml
```diff
@@ -37,3 +36,6 @@ winapi = "~0.2"
 bench = false
 name = "crust_peer"
 path = "examples/crust_peer.rs"
+
+[features]
+cargo-clippy = []
```
ustulation: Are we not able to get rid of this like in routing travis ?

.travis.yml
```diff
@@ -17,16 +17,28 @@ cache:
...
   - if [ "${TRAVIS_RUST_VERSION}" = stable ]; then
         (
             set -x;
             cargo fmt -- --write-mode=diff &&
             cargo test --verbose --release
         );
+    elif [ "${TRAVIS_OS_NAME}" = linux ]; then
+        (
+            set -x;
+            cargo clippy --features=cargo-clippy &&
```
ustulation: Do we need to provide the features=cargo-clippy ? Routing travis doesn't seem to.

Cargo.toml
vinipsmaker: The feature is needed to only disable `unknown_lints` in clippy mode. If `unknown_lints` isn't disabled, build will fail in `clippy` mode due to ` +#![cfg_attr(feature="cargo-clippy", allow(too_many_arguments))] `, which isn't a "native Rust lint".

We can only remove this if we change crust signature of some functions: http://pastebin.com/P9H6D1N1

.travis.yml
vinipsmaker: Answered in other comment

PR #812, opened 2017-01-25
Port to Mio 6

ustulation:


ustulation:
There are bugs in MIO and also changes in behavior which needs crust to handle things separately - so closing the current one as there are some critical changes to be made.

PR #822, opened 2017-03-02
test/main: add tests for connections to ourselves

nbaksalyar:


canndrew:
I'm glad to see a major open source project finally introducing tests like these.  When we don't have a connection to ourselves we lack the ability to truly connect with others, and if we stay in this state too long it severs the karmic flow through which we share in the universe's conciousness.  This, the ur-conciousness from which all conciousness (and even love itself) flows can be the only source of our spiritual fulfillment since, after all, we are the universe and it is us. For those who take a holistic world-view and appreciate the fundamental interconnectedness of all life this may seem obvious, but many software users have never tuned in to the ebbs and flows of their chakra energy fields and remain unevolved in their understanding of the dharmic mind. Only through a global awakening can we hope to overcome the massive amount of suffering in the world. This awakening must happen one concious being at a time, so thankyou MaidSafe for doing your small part.

How do we intend to reach out to and help the user when when these tests fail?

ustulation:
@canndrew well-done, you have successfully mocked my comprehension skills with that :D - anyway going to merge this now :)

PR #833, opened 2017-03-21
chore/ci: update rustc and clippy versions

ustulation:


.travis.yml
```diff
@@ -8,7 +8,7 @@ os:
 language: rust
 rust:
   - stable
-  - nightly-2016-12-19
+  - nightly-2016-03-16
```
Fraser999: Should be 2017 I think.

PR #845, opened 2017-06-20
feat/service: get peer's ip address

maqi:


src/main/service.rs
```diff
@@ -196,6 +196,11 @@ impl<UID: Uid> Service<UID> {
         }
     }
 
+    /// Return the ip address of the peer.
+    pub fn get_peer_ip_addr(&self, peer_id: &UID) -> ::Res<IpAddr> {
+        Ok(unwrap!(self.get_peer_socket_addr(peer_id)).ip())
```
ustulation: why are we using unwrap in production code when it's not used anywhere else in the call to `get_peer_socket_addr()` - why not simply `self.get_peer_socket_addr(...).map(|s| s.ip())` ? ( <-- if using that then pls copy exactly that as it's a convention elsewhere too to call socket-addr `s` inside closures)

src/main/service.rs
```diff
@@ -196,6 +196,11 @@ impl<UID: Uid> Service<UID> {
...
+    }
+
     /// Check if the provided peer address is whitelisted in config.
     pub fn is_peer_whitelisted(&self, peer_id: &UID) -> bool {
```
ustulation: Can you pls scan this file and change `peer_id` to `peer_uid` as it's done in most places - seems this file has a few places where it's missed. Do look out for changes carefully though to not shadow an existing variable named peer_uid somewhere inside the function body.

Also it seems that only these two - [this](https://github.com/maqi/crust/blob/0546dd3c379d0bf70aca0d4586bdc7a71623b4d3/src/main/service.rs#L403) and [this](https://github.com/maqi/crust/blob/0546dd3c379d0bf70aca0d4586bdc7a71623b4d3/src/main/service.rs#L389)

take `peer_uid` by value whereas the rest take by reference. Since only reference seems to be used inside the funtion body could you pls also change those to take ref `peer_uid: &UID`. (Update the call sites and your routing PR too ofc)

PR #852, opened 2017-07-11
feat/config: add dev config options

nbaksalyar:
Add dev configuration option to disable external reachability requirement.

### DIFF
--- a/installer/sample.config
+++ b/installer/sample.config
@@ -6,5 +6,6 @@
   "force_acceptor_port_in_ext_ep": false,
   "service_discovery_port": null,
   "bootstrap_cache_name": null,
-  "network_name": null
+  "network_name": null,
+  "dev": null
 }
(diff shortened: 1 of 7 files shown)


src/main/service.rs
```diff
@@ -280,11 +280,22 @@ impl<UID: Uid> Service<UID> {
...
-                        .collect(),
+                let require_external_reachability =
+                    unwrap!(self.config.lock())
+                        .cfg
+                        .dev
+                        .clone()
+                        .map_or(true,
+                                |dev_cfg| !dev_cfg.disable_external_reachability_requirement);
+                if !require_external_reachability {
+                    ExternalReachability::NotRequired
```
ustulation: This isn't correct. The sender not sending the external-reachability check does not really mean anything - it is like a client side stuff. It the rx (the network) side that needs to decide - so the code needs to be in `connection_listener/exchange_msg.rs`.

Also if someone sends ext-reachability not needed the rx (network) will put it into the bucket of a client (instead of a node) and most rpc after that will fail. If there was a test case to set the `dev` option thus, it would have caught it.

PR #854, opened 2017-08-21
test/bootstrap: fix failing bootstrap tests

canndrew:
These tests were failing (somewhat intermittently) on my machine. Make sure we don't start bootstrapping until after we've started listening.


canndrew:

PR #855, opened 2017-08-22
refactor/config: refactor config types/files

canndrew:
This gathers a bunch of config-handling stuff which was spread across multiple files into one place, changes some of the types to have more descriptive names, makes their APIs a bit more sane, and makes it possible to write tests with multiple crust instances with on-disc config files.


ustulation:
Moving config to common may not be the right thing. Common is only meant for things that are needed by more than one (isolated top level) modules (barring tests). So pls retain config defs within `main` module

src/common/config.rs.bk
```diff
@@ -0,0 +1,233 @@
+// Copyright 2017 MaidSafe.net limited.
```
ustulation: either configure gitignore or better, rustfmt with write-mode as overwrite else it might be creating these backup files by default

src/main/config_refresher.rs
```diff
@@ -78,23 +78,22 @@ impl<UID: Uid> State for ConfigRefresher<UID> {
                 }
             };
 
-        let config = match read_config_file() {
-            Ok(cfg) => cfg,
+        match self.config.reload_and_check_modified() {
+            Ok(true) => (),
+            Ok(false) => return,
```
ustulation: I might be misunderstanding code, but i think the overall logic is now incorrect. So if there was an update to config file and then someone bootstrapped off us, in connection-listener/exchange-msg, we would have reloaded the config file from the disk. Assume the new config has added the bootstrapper to the whitelist and has removed some other peer IP from the whitelist. Also say this removed peer was already an ActiveConnection. Now when the above condition is evaluated it's going to return `Ok(false)` because the in-mem and on-disc configs are the same. So the ActiveConnection which is no longer in whitelist will not be removed and if no other events occur, it will continue to remain so indefinitely. This can be an attack-vector.

In the existing code in master all these are taken care of and code would continue below to eject the ActiveConnection.

src/common/config.rs.bk
canndrew: D'oh, sorry.

canndrew:

PR #856, opened 2017-08-23
Refresh config using file-system notifications

canndrew:
Rather than periodically polling the config file.

This is based off of #855


canndrew:

PR #857, opened 2017-08-30
Core refactor

canndrew:
Switch to using a tokio core internally so that we can run futures on it. Use `PollEvented` to connect our existing mio stuff to tokio.


canndrew:

PR #864, opened 2017-10-23
Build `dev` branch on CI

povilasb:
Also, do we really want to restrict which brances to build
```
branches:
  only:
    - master
    - dev
```
or should we just build all of them?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ rust:
   - 1.19.0
   - nightly-2017-07-20
 sudo: false
-branches:
-  only:
-    - master
 cache:
   cargo: true
 before_script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -13,6 +13,7 @@ sudo: false
 branches:
```
ustulation: why not remove this completely so that it builds for everything upstream by default. Then if we ever have a brnach we dont want to build we can do something explicit then ?

Same for appveyor below.

But do check this (not specifying anything) works

povilasb: agreed :)

povilasb:
amended the commit

PR #867, opened 2017-10-24
Reenable warnings

canndrew:
I had some warnings disabled while refactoring. This reenables them.

### DIFF
--- a/src/net/peer/connect/bootstrap/cache.rs
+++ b/src/net/peer/connect/bootstrap/cache.rs
@@ -16,7 +16,6 @@
 // relating to use of the SAFE Network Software.
 
 use config_file_handler::{self, FileHandler};
-use std::ffi::OsString;
 use priv_prelude::*;
 
 pub struct Cache {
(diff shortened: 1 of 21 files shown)


src/net/peer/connect/bootstrap/cache.rs
```diff
@@ -43,6 +42,6 @@ impl Cache {
         self.file_handler.read_file().ok().unwrap_or_else(|| vec![])
     }
 
-    pub fn remove_peer_acceptor(&mut self, _peer: SocketAddr) {}
+    //pub fn remove_peer_acceptor(&mut self, _peer: SocketAddr) {}
```
povilasb: Why not just remove instead of commeting?

src/service.rs
```diff
@@ -54,7 +54,8 @@ impl<UID: Uid> Service<UID> {
         };
         MappingContext::new(options)
         .map_err(|e| CrustError::NatError(e))
-        .map(move |mc| {
+        .map(move |mut mc| {
+            mc.add_peer_stuns(config.read().hard_coded_contacts.iter().cloned());
```
povilasb: This commit seems like an important one and I guess it's related to warnings.
It would have been nice, if it was comited separately.
Just a note for the future :)

src/net/peer/connect/bootstrap/cache.rs
canndrew: I've removed it now.

src/service.rs
canndrew: You're right, that was just laziness.

canndrew:
I've fixed a nit and rebased it off of dev.

PR #868, opened 2017-10-24
fix/config: Fix config test and file watcher

canndrew:
File watcher now watches the correct location.
Config test should now pass.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -41,12 +41,13 @@ impl ConfigFile {
 
         let (tx, rx) = std::sync::mpsc::channel();
         let mut watcher = notify::watcher(tx, Duration::from_secs(1))?;
+
+        let (config_wrapper, real_path) = ConfigWrapper::open(file_name)?;
         watcher.watch(
-            &file_name,
+            &real_path,
             notify::RecursiveMode::NonRecursive,
         )?;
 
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -313,10 +314,11 @@ impl Default for ConfigSettings {
 
 impl ConfigSettings {
     /// Open the given file name and read settings.
-    pub fn open(file_name: &Path) -> Result<ConfigSettings, CrustError> {
+    pub fn open(file_name: &Path) -> Result<(ConfigSettings, PathBuf), CrustError> {
```
povilasb: A comment about returned path would be awesome.

src/config.rs
```diff
@@ -230,13 +231,13 @@ struct ConfigWrapper {
 }
 
 impl ConfigWrapper {
-    pub fn open(file_name: PathBuf) -> Result<ConfigWrapper, CrustError> {
-        let config = ConfigSettings::open(&file_name)?;
-        Ok(ConfigWrapper {
+    pub fn open(file_name: PathBuf) -> Result<(ConfigWrapper, PathBuf), CrustError> {
```
povilasb: A comment about returned path would be awesome.

PR #869, opened 2017-10-24
Build examples in CI

povilasb:
We do like to have working examples, don't we? :)


povilasb:
The build still fails because of "undeclared fs module" in tests.
So, we need to wait till that get's merged first.

PR #877, opened 2017-10-27
Enhance Service and Peer docs

povilasb:


src/service.rs
```diff
@@ -25,6 +25,19 @@ pub const SERVICE_DISCOVERY_DEFAULT_PORT: u16 = 5484;
...
+/// Use `Service` to accept incoming connections and to connect to remote peers.
+/// There are two methods to initialize connections: `connect()` and `bootstrap()`.
+/// Call `start_listener()` to accept connections.
+///
+/// By default no one can bootstrap off the listener, meaning no one can
+/// connect to you with `bootstrap()`. To enable bootstrapping call
+/// `bootstrap_acceptor()`.
+///
+/// `Service` provides futures based API, hence you can use [Tokio](https://tokio.rs/)
+/// or any other event loop compatible with [futures](https://docs.rs/futures/).
```
canndrew: You need tokio specifically because the timers and tcp sockets talk specifically to the tokio event loop.

povilasb: ok, that was a wild guess - will fix this :)

PR #878, opened 2017-10-27
Fix clippy warnings

povilasb:
Let's make clippy happy.

### DIFF
--- a/src/net/nat/hole_punch.rs
+++ b/src/net/nat/hole_punch.rs
@@ -20,6 +20,7 @@ use util;
 
 /// Punch a hole to a remote peer. Both peers call this simultaneously try to perform a TCP
 /// rendezvous connect to each other.
+#[cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]
 pub fn tcp_hole_punch(
     handle: &Handle,
     socket: TcpBuilder,
--- a/src/net/peer/connect/bootstrap/bootstrap.rs
+++ b/src/net/peer/connect/bootstrap/bootstrap.rs
@@ -64,7 +64,7 @@ pub fn bootstrap<UID: Uid>(
     ext_reachability: ExternalReachability,
     blacklist: HashSet<SocketAddr>,
     use_service_discovery: bool,
-    config: ConfigFile,
+    config: &ConfigFile,
 ) -> BoxFuture<Peer<UID>, BootstrapError> {
     let handle = handle.clone();
     let try = || -> Result<_, BootstrapError> {
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,1 @@
+doc-valid-idents = ["UPnP"]
--- a/examples/exchange_data.rs
+++ b/examples/exchange_data.rs
@@ -20,8 +20,8 @@
 //! In a nutshell all it does is:
 //!
 //! 1. connects to remote peer (for connection details see `connect.rs` sample)
-//! 2. sends "Hello from peer '$peer_id'" message
-//! 3. sends "Goodbye from peer '$peer_id" message
+//! 2. sends "Hello from peer '`$peer_id`'" message
+//! 3. sends "Goodbye from peer '`$peer_id`'" message
 //! 4. prints any messages received from remote peer
 
 #[macro_use]
--- a/examples/exchange_data.rs
+++ b/examples/exchange_data.rs
@@ -53,7 +53,7 @@ fn main() {
     println!("Service id: {}", service_id);
 
     // does the p2p connection as demonstrated in `connect.rs` example
-    let peer = connect_to_peer(&mut event_loop, service_id.clone());
+    let peer = connect_to_peer(&mut event_loop, service_id);
     println!(
         "Connected to peer: {}, {}",
         peer.uid(),
--- a/src/compat/service.rs
+++ b/src/compat/service.rs
@@ -77,8 +77,8 @@ impl<UID: Uid> Service<UID> {
                     if node_ips.is_none() && client_ips.is_none() {
                         return Ok(());
                     }
-                    let node_ips = node_ips.unwrap_or(HashSet::new());
-                    let client_ips = client_ips.unwrap_or(HashSet::new());
+                    let node_ips = node_ips.unwrap_or_default();
+                    let client_ips = client_ips.unwrap_or_default();
                     cm.whitelist_filter(client_ips, node_ips);
                     Ok(())
                 })
--- a/src/compat/service.rs
+++ b/src/compat/service.rs
@@ -421,14 +421,14 @@ impl<UID: Uid> Service<UID> {
                     ::service::SERVICE_DISCOVERY_DEFAULT_PORT,
                 );
                 let f = {
-                    service_discovery::discover::<Vec<SocketAddr>>(&handle, sd_port)
+                    service_discovery::discover::<Vec<SocketAddr>>(handle, sd_port)
                         .into_future()
                         .map(|s| s.infallible())
                         .flatten_stream()
                         .first_ok()
                         .map(|_| ())
                         .map_err(|_| ())
-                        .with_timeout(&handle, Duration::from_secs(1), ())
+                        .with_timeout(handle, Duration::from_secs(1), ())
                         .then(move |res| {
                             let _ = tx.send(res.is_ok());
                             Ok(())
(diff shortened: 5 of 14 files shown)


src/net/nat/mapped_tcp_socket.rs
```diff
@@ -40,7 +40,7 @@ quick_error! {
 }
 
 /// Returns a TCP socket and a set of possible external addresses of the socket. For instance, if
-/// we're behind a NAT then this will use UPnP and a pseudo-STUN protocol to try and find global IP
+/// we're behind a NAT then this will use `UPnP` and a pseudo-STUN protocol to try and find global IP
```
canndrew: hmm, why did clippy do that :/

povilasb: good question. Btw, this is the claimed issue: https://rust-lang-nursery.github.io/rust-clippy/v0.0.166/index.html#doc_markdown

povilasb: I could research this issue further and maybe even raise clippy issue?

canndrew: According to that page we should be able to add UPnP to `doc_valid_idents`.

src/net/nat/hole_punch.rs
```diff
@@ -22,7 +22,7 @@ use util;
 /// rendezvous connect to each other.
 pub fn tcp_hole_punch(
     handle: &Handle,
-    socket: TcpBuilder,
+    socket: &TcpBuilder,
```
canndrew: You can only actually "use" a `TcpBuilder` (via `listen` or `connect`) once, so it's a shame it wants to make this change. It makes more sense for this function to consume the builder.

It's too minor an issue to argue with clippy over though, so I'm happy to merge this.

povilasb: no problem, I'll revert the change and disable warning with `#[cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]`

src/net/peer/connect/bootstrap/bootstrap.rs
```diff
@@ -64,7 +64,7 @@ pub fn bootstrap<UID: Uid>(
     ext_reachability: ExternalReachability,
     blacklist: HashSet<SocketAddr>,
     use_service_discovery: bool,
-    config: ConfigFile,
+    config: &ConfigFile,
```
ustulation: When changing everything to reference, also keep in mind that "Copy" types are better passed by value (and hence copied) - at-least if it's not huge arrays inside structs etc. Most of the time they will be NRVO optimised out unlike references which make it relatively more opaque to optimisations depending on scenarios. Don't know if it applies to rust, but certainly was something i remember from C/C++ days (and i have forgotten most of it anyway). Refer to literature like [these](https://stackoverflow.com/a/40185996/1060004) and NRVOs etc.

Could you pls scan and confirm this and correct it if necessary ?

povilasb:
> (1) Ah we don't use 72eaf2c#diff-acc964cd562925049c345acfaf318e92R24 anywhere. So explicitly specify the stuff you are to use - never wildcards.

Do you mean wildcard imports `use priv_prelude::*;`?
If so, this is out of this PR scope, which is only meant to fix clippy issues.
But I will gladly make another refactoring PR. This Pr is already 12 commits long and I'd like to keep them shorter anyway to make it easier to review and revert if necessary...

src/net/peer/connect/bootstrap/bootstrap.rs
povilasb: Well, this seems like a tricky question.
I couldn't find a definite answer regarding NVO in Rust. 
Some people claim it's there (LLVM might be doing that): https://stackoverflow.com/a/27837182/3323667.
I also seems like some related work is still in progress: https://github.com/rust-lang/rust/issues/32966.

I was following `clippy` lint https://rust-lang-nursery.github.io/rust-clippy/v0.0.166/index.html#needless_pass_by_value:
```
warning: this argument is passed by value, but not consumed in the function body
  --> src/net/peer/connect/bootstrap/bootstrap.rs:67:13
   |
67 |     config: ConfigFile,
   |             ^^^^^^^^^^ help: consider taking a reference instead: `&ConfigFile`
   |
   = note: #[warn(needless_pass_by_value)] on by default
   = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.166/index.html#needless_pass_by_value
```

And I still don't know how using reference might be worse than consuming a value.

ustulation: It's worse because of the reasons stated. Pointers are bad for stuffs like cache-locality etc. Also if taking something as `var: const i32*` will certainly lead to more code generation than `var: i32` due to load semantics. Since references are pointers behind the scene, same applies to them.

The reason why clippy complains here is because [ConfigFile is non-copy type](https://github.com/maidsafe/crust/blob/e77e71cf7f96e60c41c5403d72558f52d68de8e6/src/config.rs#L32), and for those things you can only move, clone or pass reference. In this case it makes sense to pass reference. That is why the review comment explicitly mentioned
> ... "Copy" types are better passed by value ...

Anyway do keep a note to make sure that we are not always using references but checking if it's a Copy type and then evaluating it what should we use (which was the point of the review comment - to scan and check we are just not rampantly doing it for everything).

povilasb: yeah, it's `tests::compat_api::start_two_services_exchange_data` which hangs.
It already does in `dev` branch, so I'd like to research and fix it with another PR.

povilasb: also, `ConfigFile` itself is `Arc`
```
/// A handle to a crust config file. This handle can be cloned and shared throughout the program.
#[derive(Clone)]
pub struct ConfigFile {
    inner: Arc<RwLock<ConfigWrapper>>,
}
```
so it's definitely no point of referencing to it.

ustulation: Referencing is better than cloning - cloning will unnecessary cause an instruction emission for atomic increase of reference count, which in this case, being and Arc instead of Rc, will be atomic (which is certainly not a trivial as for Rc). Then the local destruction will cause another one to reduce the atomic reference count, which is again not optimal. Either or both of these would be sequentially consistent and thus interfere with optimisers. Anything that involves thread safety is not really optimal and hence should be used only if thread safety is required, else it's like having a mutex locked and unlocked in a single threaded system - buys nothing but wastes cpu cycles and prevents optimisations (they are like memory (re-ordering) barriers).

Hence (just my opinion) Arc's should possible be passed by reference unless sending over threads where ofc you should pass a clone (and safe Rust will not allow you to do otherwise anyway).

povilasb: that makes sense :)

canndrew:
> (1) Ah we don't use 72eaf2c#diff-acc964cd562925049c345acfaf318e92R24 anywhere. So explicitly specify the stuff you are to use - never wildcards.

Any chance we could amend this rule to allow this specific pattern?

I understand the motivation for banning wildcard imports - and I agree with it - they make code hard to read and hard to explore because you can't tell where things are being imported from. You can ameliorate these problems though if you:

 (a) Only use a single wildcard import per module.
 (b) Use the same wildcard import throughout every module in the crate.

That's what I've done in crust. There's a `priv_prelude` module in the crate root which gets `use priv_prelude::*;`-ed into every other module. Nothing else ever gets wildcard-imported, so if you want to know where something is defined, and it isn't being explicitly imported, you just need to look in `priv_prelude.rs`. I did this because the same bunch of things get used in loads of modules throughout crust. Without it, every module needs to have most of:

```rust
pub use future_utils::{BoxFuture, BoxStream, FutureExt, IoFuture, IoStream, StreamExt};
pub use futures::{Async, AsyncSink, Future, IntoFuture, Sink, Stream, future, stream};
pub use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
pub use std::net::{SocketAddr, SocketAddrV4, SocketAddrV6};
pub use std::path::{Path, PathBuf};
pub use std::time::{Duration, Instant};
etc.
etc.
etc.
```

Which gets tiring to write and is time-consuming to fix the missing import errors. It's also kinda pointless - anyone reading this is gonna know what `Future` means, we don't need to specify it separately in every file.

canndrew:
There was [an RFC](https://github.com/rust-lang/rfcs/pull/890) offering an alternative to this at one point. But it doesn't look like it's going to go anywhere anytime soon.

PR #879, opened 2017-10-31
Fix hanging test

povilasb:
During the test 200 buffers of 2 MB are randomized and exchanged between two `Crust` `Services` running on separate threads.
That takes a lot of time. Just buffer randomization consumes about 55 seconds.
I believe we don't need such big buffers to validate data exchange.
This fix reduces the buffers, hence tests don't timeout.


povilasb:
Windows build fails because of another test:
```
tests::compat_api::bootstrap_using_hard_coded_contacts ... FAILED
```
which timesout.

PR #881, opened 2017-11-03
Fix failing tests

povilasb:
All tests pass on Linux and Mac OS X, but some fail on Windows.
This PR fixes them.


src/net/socket.rs
```diff
@@ -347,7 +348,10 @@ mod test {
             f0
             .join(f1)
             .and_then(|((), ())| Ok(()))
-            .map_err(|e| panic!(e))
+            .map_err(|e| {
+                println!("Error: {:?}", e);
```
canndrew: This shouldn't be necessary since we're panicking. Also, if we want to print stuff during tests it generally better to use the log macros (`debug!()`, `error!()`, etc.)

povilasb: yes, you;re right, I'll remove this commit.
I guess I got confused by the output of `panic!`, but I just reran it and it's ok :)

PR #887, opened 2017-11-28
Use p2p for rendezvous connect

povilasb:


src/error.rs
```diff
@@ -66,9 +66,8 @@ quick_error! {
             from()
         }
         /// Error preparing connection info.
-        PrepareConnectionInfo(e: io::Error) {
```
canndrew: Why was the error removed? It looks like (futher down) the error that should be getting wrapped here just gets discarded.

povilasb: Yeah, I just remove the `io:Error parameter`, because unbonded channel doesn't produce this error: ```
        ch1.into_future()
            .and_then(move |(msg_opt, chann)| {
                Ok(PrivConnectionInfo {
                    id: our_uid,
                    for_direct: direct_addrs.into_iter().collect(),
                    for_hole_punch: vec![],
                    hole_punch_socket: None,
                    p2p_conn_info: unwrap!(msg_opt),
                    rendezvous_channel: chann,
                    connection_rx: conn_rx,
                })
            })
            .map_err(|_| CrustError::PrepareConnectionInfo)
            .into_boxed()
```

File: src/net/peer/connect/connect.rs [hunk 2]
```diff
@@ -270,3 +177,20 @@ fn validate_connect_request<UID: Uid>(
     }
     Ok(())
 }
+
+
+/// Starts p2p rendezvous connect.
+///
+/// Gets peer info from rendezvous relay channel and sends connected tcp stream via `out`
+/// channel.
+pub fn rendezvous_connect(
```
canndrew: This function was a bit confusing for me - the way it returns its result via a `oneshot` rather than just directly, and the way it returns `()` for an error. Then I realized it's because it gets turned into a background task with `handle.spawn`.

Since the only point of the future returned here is to be passed to `handle.spawn` I'd prefer it if either (a) the result being wrapped and sent through the channel happened over where `handle.spawn` gets called or (b) `handle.spawn` got called in this function.

Also the `oneshot` channel should probably carry a `Result`. There's also the `.log_error` method for use by daemon-like futures if you don't want to pass the error back.

src/error.rs
canndrew: What error is getting produced here but getting discarded? If it's a `Void` then we don't need `CrustError::PrepareConnectInfo` at all and we can just use `.infallible()`. If it's a `()` then write `.map_err(|()| CrustError::PrepareConnectionInfo)`. If it's anything else then that should be the error type carried by `PrepareConnectionInfo`.

src/net/peer/connect/connection_info.rs
```diff
@@ -24,10 +28,12 @@ pub struct PrivConnectionInfo<UID> {
...
-    #[doc(hidden)]
-    pub for_hole_punch: Vec<SocketAddr>,
-    #[doc(hidden)]
-    pub hole_punch_socket: Option<TcpBuilder>,
+    // P2P connection info is optional in case when p2p `rendezvous_connect` fails.
+    // In such cases we can still proceed with direct connections, but we must keep track
+    // that P2P connections should not be attempted.
+    pub p2p_conn_info: Option<Bytes>,
+    pub rendezvous_channel: UnboundedBiChannel<Bytes>,
+    pub connection_rx: Option<oneshot::Receiver<Result<TcpStream, SingleConnectionError>>>,
```
canndrew: It looks like `p2p_conn_info` and `connection_rx` shouldn't be two separate `Option`s. If they're always both `Some` or both `None` then they should be combined into a single `Option`. Create a new type to put in the `Option` if necessary.

src/net/peer/connect/connect.rs
```diff
@@ -81,170 +83,161 @@ quick_error! {
...
+    rendezvous_channel: UnboundedBiChannel<Bytes>,
+    conn_rx: Option<oneshot::Receiver<Result<TcpStream, SingleConnectionError>>>,
+    p2p_conn_info: Option<Vec<u8>>,
+) -> BoxFuture<TcpStream, SingleConnectionError> {
+    if p2p_conn_info.is_none() || conn_rx.is_none() {
+        return future::empty().into_boxed();
+    }
 
-    future::result(try()).flatten().into_boxed()
+    let conn_info = Bytes::from(unwrap!(p2p_conn_info));
```
canndrew: You don't need to use an `unwrap!` here. Instead of using `p2p_conn_info.is_none()` and then unwrapping, use a match statement to get the connection info and then use it directly.

More generally, never use a panicking dynamic check like `unwrap!` if you can get the same guarantee from the type system.

eg:

```rust
// BAD
if foo.is_none() {
    return;
}
let bar = unwrap!(foo);

// GOOD
let bar = match foo {
    Some(bar) => bar,
    None => return,
};
```

With the `BAD` code above, you may think you "know" that unwrapping is safe because foo is always `Some(..)`, but don't rely on knowing things when you can use the type checker to enforce them. Things that we "know" work can still break if we make a mistake, or if the code gets refactored and a bug gets introduced. Things that are checked by the type systems are guaranteed to work or will break at compile time if someone breaks the code. Also, using correct types makes code more readable - if you use an `unwrap` I have to look around and figure out why you're so sure the `Option` must contain a `Some`.

src/net/peer/connect/connect.rs
```diff
@@ -81,170 +83,161 @@ quick_error! {
...
+    ).map_err(SingleConnectionError::Io)
+        .into_boxed()
+}
 
-        let all_connections = direct_incoming.select(other_connections);
+fn handshake_incoming_connections<UID: Uid>(
+    our_connect_request: ConnectRequest<UID>,
+    conn_rx: UnboundedReceiver<ConnectMessage<UID>>,
+    their_id: UID,
+    name_hash: NameHash,
```
canndrew: If we've already checked that this is the same `NameHash` as in `our_connect_request` then we don't need to take it as a separate argument. Same with `handle_outgoing_connections`.

src/net/peer/connect/connect.rs
```diff
@@ -81,170 +83,161 @@ quick_error! {
...
+) -> BoxStream<(Socket<HandshakeMessage<UID>>, UID), SingleConnectionError> {
+    conn_rx
+        .map_err(|()| unreachable!())
```
canndrew: I'd prefer to use the `Unbounded{Sender,Reciever}` from `future-utils` just so we don't have to have `unreachable!`s in our code. Although it's a shame not use a more standard type.

src/net/peer/connect/connect.rs
```diff
@@ -81,170 +83,161 @@ quick_error! {
         TimedOut {
             description("connection attempt timed out")
         }
+        DeadChannel {
+            description("Communication channel was cancelled")
+        }
+        RendezvousConnect {
```
canndrew: This should still carry the error from where it's used.

src/net/peer/connect/connect.rs
```diff
@@ -267,3 +260,24 @@ fn validate_connect_request<UID: Uid>(
...
+/// connection receiver
+pub fn start_rendezvous_connect(
+    handle: &Handle,
+    rendezvous_relay: UnboundedBiChannel<Bytes>,
+) -> oneshot::Receiver<Result<TcpStream, SingleConnectionError>> {
+    let (conn_tx, conn_rx) = oneshot::channel();
+    let start_conn = TcpStream::rendezvous_connect(rendezvous_relay, handle)
+        .map_err(|_| SingleConnectionError::RendezvousConnect)
+        .then(move |result| conn_tx.send(result))
+        .then(|_| Ok(()));
```
canndrew: Small nit: could this be replaced with a `map_err`? Just to make it clearer that there isn't an `Ok` value of some kind being thrown away.

src/net/peer/connect/connect.rs
povilasb: Yep, this parameter is redundant - result of blind refactoring, will fix this.

src/net/peer/connect/connect.rs
povilasb: Hmm, how do I do that?
Sometimes it's https://docs.rs/future-utils/0.5.0/future_utils/mpsc/struct.SendError.html and sometimes it's https://docs.rs/futures/0.1.17/futures/sync/oneshot/struct.Canceled.html

But they both basically mean the same thing: the other side of synchronization channel was dropped.

canndrew: I can only find `RendezvousConnect` being created in one place: https://github.com/maidsafe/crust/pull/887/files#diff-56c4cb5be4a0c20c65408095750c183eR278

povilasb: oh, I though you were about `DeadChannel` error. Never mind then :)

povilasb: Hmm, `RendezvousConnectError` is generic this applies https://github.com/tailhook/quick-error/issues/20
Currently, I changed it to 
`  RendezvousConnect(e: TcpRendezvousConnectError<UnboundedBiChannel<Bytes>>) {`

but that looks quite fragile, doesn't it?

src/net/peer/connect/connect.rs
povilasb: Well, that's the signature required by [Handle::spawn()](https://docs.rs/tokio-core/0.1.10/tokio_core/reactor/struct.Handle.html#method.spawn)

The only possible error in this case could be failure to send: https://docs.rs/futures/0.1.17/futures/sync/oneshot/struct.Sender.html#method.send
And this would only happen if the receiving part (someone who's interested in connection) got closed. In such case, they only way to handle this is do nothing. Hence, forget about the result.

src/net/peer/connect/connect.rs
povilasb: Makes sense, I'd agree with that. Except this requires more changes in different parts of code (`demux`) since `conn_rx` is passed to us by `demux`.
Anyway, I believe this change is out of scope of this PR.

PR #892, opened 2017-12-04
Fix connect

povilasb:
When there are multiple direct connections, some of them prematurely die.

### DIFF
--- a/src/net/peer/connect/connect.rs
+++ b/src/net/peer/connect/connect.rs
@@ -117,7 +117,7 @@ pub fn connect<UID: Uid>(
 
     let direct_connections = connect_directly(handle, their_info.for_direct);
     let p2p_connection = connect_p2p(our_info.p2p_conn_info, their_info.p2p_conn_info);
-    let all_connections = handshake_outgoing_connections(
+    let all_outgoing_connections = handshake_outgoing_connections(
         handle,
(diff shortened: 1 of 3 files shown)


src/tests/compat_api.rs
```diff
@@ -232,7 +232,10 @@ fn start_two_services_exchange_data() {
     });
 
     let service0 = j0.join();
+    assert!(service0.is_ok());
```
canndrew: If you use `unwrap!(service0)` here that'll print the error.

src/net/peer/connect/connect.rs
```diff
@@ -119,24 +119,75 @@ pub fn connect<UID: Uid>(
...
 
     let direct_incoming = handshake_incoming_connections(our_connect_request, peer_rx, their_id);
+    let all_connections = all_outgoing_connections.select(direct_incoming);
+    choose_peer(&handle, all_connections, our_info.id, their_id)
+}
+
+/// Takes all pending handshaken connections and chooses the first one successful.
+///
+/// Then this connection is wrapped in a `Peer` structure.
+/// Depending on service id either initiates connection choise message or waits for one.
```
canndrew: typo "choise"

ustulation:
OSx wasn't starting in CI

PR #893, opened 2017-12-04
Don't filter loopback addrs in Listener

canndrew:
These already get filtered later on. If we filter them in here then we won't be able to connect over loopback (and also it breaks a test).

### DIFF
--- a/src/net/listener.rs
+++ b/src/net/listener.rs
@@ -144,11 +144,7 @@ fn listener_addresses(
     public_addr: Option<SocketAddr>,
 ) -> io::Result<(TcpListener, HashSet<SocketAddr>)> {
     listener.expanded_local_addrs().map(|local_addrs| {
-        let mut addrs = local_addrs
-            .iter()
-            .filter(|addr| !addr.ip().is_loopback())
-            .cloned()
-            .collect::<HashSet<SocketAddr>>();
+        let mut addrs = local_addrs.iter().cloned().collect::<HashSet<SocketAddr>>();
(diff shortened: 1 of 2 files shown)


povilasb:
what about direct connections? I guess you're referring to `p2p` when you say "they get filtered later on".
Also, I remember some discussions with David that we don't really want to connect to localhost because that might result in connecting to ourselves, no?

canndrew:
There's a function in `p2p` called `filter_addrs` which will filter not just loopback but also local network addresses like 192.168.*.*. But it will only filter addresses that we know are useless based on what our full set of IP addresses are compared to the peer's. So this function should be used instead when doing a direct connect. It doesn't make sense to filter them in `Listener` since we are in-fact listening on loopback. It's not `Listener`'s job to decide what addresses other parts of the code might find useful, just to describe what addresses it has.

Also any address can be used to connect to ourself, there's nothing special about loopback in that regard. We already have checks in place which discard those connections if they happen.

povilasb:
Makes sense, then I agree with you.

Anyway, what about public conn info: is localhost address supposed to be included in there?

canndrew:
> Anyway, what about public conn info: is localhost address supposed to be included in there?

Sure, include all the addresses. `filter_addrs` will filter out any addresses it can't use. The only time it will ever not filter a loopback is if neither peer can detect their global IP address or both have the same global IP and also neither peer have a local network address or both have the same local network address. In other words: if there's good reason to think we're running on the same machine or if we have no other addresses and we're desperate.

PR #894, opened 2017-12-05
use STUN

povilasb:
Uses `hard_coded_contacts` config value as STUN server list.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,6 +15,8 @@
 // Please review the Licences for the specific language governing permissions and limitations
 // relating to use of the SAFE Network Software.
 
+#![forbid(warnings)]
+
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -30,7 +30,7 @@ script:
     elif [ "${TRAVIS_OS_NAME}" = linux ]; then
       (
         set -x;
-        cargo clippy && cargo clippy --profile=test
+        cargo clippy && cargo clippy --profile=test -- -D warnings
```
ustulation: Use linters from master. This should not be required then.

povilasb:
I tried to add all lint rules from master branch, but then bunch of errors poured in.
So I'll leave that for the next PR.

PR #896, opened 2017-12-07
Replace `get_if_addrs` module with `get_if_addrs` crate.

madadam:
Fixes build issues on some mobile platforms.


canndrew:
@madadam The `get_if_addrs.rs` file can be removed.

PR #897, opened 2017-12-08
Utp integration

canndrew:


src/net/mod.rs
```diff
@@ -22,8 +22,11 @@ pub use self::peer::{Acceptor, BootstrapAcceptError, BootstrapAcceptor, Bootstra
                      SingleConnectionError, Uid, bootstrap};
 pub use self::service_discovery::ServiceDiscovery;
 pub use self::socket::{MAX_PAYLOAD_SIZE, Priority, Socket, SocketError};
+pub use self::protocol_agnostic::{PaAddr, PaStream, PaListener};
```
povilasb: Isn't `Pa` prefix redundant, since all of them (`PaAddr`, `PaStream`, etc.) belong to the same module `protocol_agnostic`?

Cargo.toml
```diff
@@ -41,6 +41,7 @@ tiny-keccak = "1.3"
 tokio-core = "0.1.10"
 tokio-io = "0.1.3"
 unwrap = "1.1"
+url = "1.5.1"
```
povilasb: I believe we have decided to use ~ requirements for versions: http://doc.crates.io/specifying-dependencies.html#tilde-requirements 
hence `url = "~1.5.1` in this case
Also, see this: https://github.com/maidsafe/routing/blob/master/Cargo.toml

src/net/protocol_agnostic/addr.rs
```diff
@@ -1,6 +1,91 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use priv_prelude::*;
+    use std::str::FromStr;
+
+    #[test]
+    fn test() {
```
povilasb: Test case description would be nice: in case it fails, it would provide more details what's wrong, .e.g.
```
#[test]
fn from_str_parses_address_from_url()
```

src/net/protocol_agnostic/addr.rs
```diff
@@ -1,6 +1,91 @@
...
+    type Err = ParseError;
+
+    fn from_str(s: &str) -> Result<PaAddr, ParseError> {
+        let url = match Url::parse(s) {
+            Ok(url) => url,
+            Err(e) => return Err(ParseError::MalformedUrl(e)),
+        };
+        let ret = match url.scheme() {
+            "tcp" => {
+                let ip = match url.host_str() {
```
povilasb: It would be nice to move the body of this arm to a separate function: it would be easier to test and `from_str` would be more readable and extendible. E.g. 
```
let ret = match url.scheme() {
    "tcp" => parse_tcp_addr(url),
    "udp" => parse_udp_addr(url),
    _ => Err(...),
}
```

src/net/protocol_agnostic/addr.rs
```diff
@@ -49,20 +53,25 @@ impl FromStr for PaAddr {
...
-                };
-                let port = match url.port() {
-                    Some(port) => port,
-                    None => return Err(ParseError::MissingPort),
-                };
-                let addr = SocketAddr::new(ip, port);
+                PaAddr::Tcp(addr)
+            },
+            "utp" => {
                 PaAddr::Tcp(addr)
```
povilasb: I guess that was supposed to be `PaAddr::Utp` :)

src/net/mod.rs
canndrew: Yeah but then they get used in other places throughout crust I didn't want to have to write `protocol_agnostic::` everywhere. Also this naming is consistent with `TcpStream`, `TcpListener`, `UtpStream`, `UtpListener` etc.

povilasb: true

src/net/protocol_agnostic/msg.rs
```diff
@@ -0,0 +1,8 @@
+use priv_prelude::*;
```
povilasb: Missing MaidSafe license header.

src/net/protocol_agnostic/msg.rs
```diff
@@ -0,0 +1,8 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct PaRendezvousMsg {
```
povilasb: Why  not move `PaRendezvousMsg` to `protocol_agnostic/mod.rs`?
Currently, it's an 8 line file and it also has redundant labeling: `msg::PaRendezvousMsg`.

src/net/protocol_agnostic/stream.rs
```diff
@@ -36,6 +44,124 @@ impl PaStream {
         }
     }
 
+    pub fn rendezvous_connect<C>(
```
povilasb: That is quite complex function with a lot of logic and branching - would you care to decompose it?

src/compat/event_loop.rs
```diff
@@ -65,10 +65,13 @@ pub fn spawn_event_loop<UID: Uid>(
...
         match try() {
             Ok((mut core, mut service_state)) => {
+                let handle = core.handle();
                 let (tx, rx) = mpsc::unbounded::<ServiceCommand<UID>>();
                 unwrap!(result_tx.send(Ok(tx)));
                 unwrap!(core.run({
-                    rx.for_each(move |cb| Ok(cb.call_box(&mut service_state)))
+                    rx
+                    .for_each(move |cb| Ok(cb.call_box(&mut service_state)))
+                    .and_then(move |()| Timeout::new(Duration::from_millis(200), &handle).infallible())
```
povilasb: Just curious, what does that buy us?

canndrew: It inelegantly fixes the bug which I think was causing `compat_api::bootstrap_using_hard_coded_contacts` to fail. It you close a uTP connection then immediately shutdown the event loop it's possible for the FIN packet to not get sent. Keeping the event loop alive a little while longer gives some time for the socket to flush itself.

src/net/peer/connect/bootstrap/mod.rs
```diff
@@ -16,8 +16,8 @@
 // relating to use of the SAFE Network Software.
 
 mod cache;
-mod bootstrap;
+mod bootstrap_fn;
```
povilasb: Naming module after function (`_fn`) sounds very specific, doesn't it?
What if we simply moved `bootstrap()` function out of `bootstrap` module to `mod.rs`?

src/net/service_discovery/mod.rs
```diff
@@ -18,11 +18,11 @@
 mod msg;
 mod server;
 mod discover;
-mod service_discovery;
+mod service_discovery_ty;
```
povilasb: What does `_ty` stand for?

src/net/peer/connect/bootstrap/mod.rs
canndrew: Ugh, we could. I just prefer to keep code in leaf modules rather than putting it in `mod.rs`. I don't like the naming here either though.

PR #899, opened 2017-12-14
feature/use_local_port: Enable p2p's force_use_local_port feature

canndrew:


canndrew:
This was included in #901

PR #900, opened 2017-12-14
fix/whitelisting: respect whitelists in rendezvous connect

canndrew:


canndrew:
This was included in #901

PR #901, opened 2017-12-14
fix/utp: switch to tokio-utp-0.2.9

canndrew:
tokio-utp-0.2.9 contains a bugfix


canndrew:
This PR now contains #899 and #900

PR #903, opened 2017-12-19
enable disabling tcp for testing

canndrew:


src/config.rs
```diff
@@ -245,7 +245,7 @@ impl ConfigWrapper {
         ))
     }
 
-    pub fn reload(&mut self) -> Result<(), CrustError> {
+    fn reload(&mut self) -> Result<(), CrustError> {
```
povilasb: It seems like making these methods private doesn't really have anything to do with TCP disable. Hence, I would put those changes to a separate commit and keep my commits atomic. The benefits of that is explained here: https://www.freshconsulting.com/atomic-commits/

src/net/protocol_agnostic/stream.rs
```diff
@@ -77,20 +92,25 @@ impl PaStream {
...
                 .map_err(|(v, _)| void::unreachable(v))
-                .and_then(|(tcp_msg_opt, tcp_ch_0)| {
+                .and_then(move |(tcp_msg_opt, tcp_ch_0)| {
                     utp_ch_0
                         .into_future()
                         .map_err(|(v, _)| void::unreachable(v))
                         .and_then(move |(utp_msg_opt, utp_ch_0)| {
                             let msg = PaRendezvousMsg {
-                                tcp: tcp_msg_opt,
+                                tcp: if disable_tcp { None } else { tcp_msg_opt },
```
povilasb: Mmm, I so much would like to have unit tests for such code branches.

src/net/protocol_agnostic/stream.rs
```diff
@@ -77,20 +92,25 @@ impl PaStream {
         <C as Sink>::SinkError: fmt::Debug,
         C: 'static,
     {
+        let disable_tcp = match config.read().dev {
```
povilasb: Duplicate code with https://github.com/maidsafe/crust/pull/903/commits/78c87256506593092e8075d6ac8e39d72d6f17df#diff-2769f48124e6f45bc5d555b5c1d61f29R55

Couldn't this be some function of config, e.g. `config.tcp_enabled() -> bool`

src/service.rs
```diff
@@ -132,9 +132,16 @@ impl<UID: Uid> Service<UID> {
     /// address, drop a `Listener` to stop listening on its address. The stream will end once all
     /// configured listeners have been returned.
     pub fn start_listening(&self) -> BoxStream<Listener, CrustError> {
+        let disable_tcp = match self.config.read().dev {
```
povilasb: Duplicate  see above

canndrew:
This is deprecated by #909

PR #907, opened 2017-12-22
Refactoring

povilasb:
More code refactoring based on automated linting report.


src/util/test.rs
```diff
@@ -43,7 +43,6 @@ pub fn random_id() -> UniqueId {
 
 pub fn random_vec(size: usize) -> Vec<u8> {
     let mut ret = Vec::with_capacity(size);
-    unsafe { ret.set_len(size) };
```
ustulation: hmm are you sure we can remove this and this would still work ? @povilasb @canndrew

povilasb: hm, you're right, just tested that manually - crashes.
Wonder how tests passed :/
The reason I did this change is that I thought `Vec::with_capacity` already does that. But reading the docs closer now I think I understand that it allocates the necessary amount of memory but keeps internal field `.length = 0`.
> It is important to note that this function does not specify the length of the returned vector, but only the capacity.

PR #909, opened 2018-01-03
Utp rendezvous connect fix

canndrew:
This is actually a whole bunch of changes, it contains:
 * tcp disabling
 * connect example changes
 * igd disabling
 * utp rendezvous connect fix

I'm PRing them together because they were built on top of each other and it's probably more effort than it's worth to tease them apart into separate PRs.


povilasb:
not all commit messages comply with our style: https://github.com/autumnai/leaf/blob/master/CONTRIBUTING.md#git-commit-guidelines

src/net/peer/connect/bootstrap/mod.rs
```diff
@@ -65,6 +65,7 @@ pub fn bootstrap<UID: Uid>(
     use_service_discovery: bool,
     config: &ConfigFile,
 ) -> BoxFuture<Peer<UID>, BootstrapError> {
+    let config = config.clone();
```
povilasb: why an extra copy, if we take a reference anyway?

src/net/peer/connect/mod.rs
```diff
@@ -124,9 +124,10 @@ pub fn connect<UID: Uid>(
     name_hash: NameHash,
     our_info: PrivConnectionInfo<UID>,
     their_info: PubConnectionInfo<UID>,
-    config: ConfigFile,
+    config: &ConfigFile,
     peer_rx: UnboundedReceiver<ConnectMessage<UID>>,
 ) -> BoxFuture<Peer<UID>, ConnectError> {
+    let config = config.clone();
```
povilasb: redundant copy?

src/net/protocol_agnostic/stream.rs
```diff
@@ -79,20 +94,25 @@ impl PaStream {
         <C as Sink>::SinkError: fmt::Debug,
         C: 'static,
     {
+        let disable_tcp = match config.read().dev {
```
povilasb: Duplicate code: https://github.com/maidsafe/crust/pull/909/commits/e88a2c8990d7bf004c4a7c18c423a11f48a64dc3#diff-2769f48124e6f45bc5d555b5c1d61f29R56
What about moving this to config class, hence smth like `if config.read().tcp_disabled()` could be possible?

src/config.rs
```diff
@@ -155,6 +155,14 @@ impl ConfigFile {
         res
     }
 
+    /// Check whether TCP is disabled for testing.
+    pub fn tcp_disabled(&self) -> bool {
```
povilasb: Hehe, if you had squashed this commit with previous one, I wouldn't have commented several things :)

src/net/protocol_agnostic/stream.rs
```diff
@@ -370,3 +370,50 @@ where
...
+
+#[cfg(test)]
+mod test {
+    use priv_prelude::*;
+    use config::DevConfigSettings;
+    use future_utils::bi_channel;
+    use tokio_core::reactor::Core;
+
+    #[test]
+    fn direct_connect_tcp_disabled() {
```
povilasb: It's also useful to spell out what's expected, e.g. 
```
#[test]
fn direct_connect_uses_other_than_tcp_addrs_when_tcp_disabled()
```
Personally, I like to scope tests:
```
mod direct_connect {
    #[test]
    fn it_uses_other_than_tcp_addrs_when_tcp_disabled {
    }
}
```

examples/connect.rs
```diff
@@ -55,12 +55,11 @@ extern crate crust;
 
 
 use crust::{ConfigFile, PubConnectionInfo, Service, Uid};
-use crust::config::DevConfigSettings;
 use futures::Stream;
 
 use futures::future::empty;
 use rand::Rng;
-use std::{fmt, io, env};
+use std::{fmt, io};
```
povilasb: Consider squashing such commits: less commits to review ;)

examples/connect.rs
```diff
@@ -78,60 +97,111 @@ impl fmt::Display for PeerId {
...
-        unwrap!(
-        event_loop.run(service.connect(our_conn_info, their_info)),
-        "Failed to connect to given peer",
-    );
-    println!("Connected to peer: {}", peer.uid());
-
-    // Run event loop forever.
-    let res = event_loop.run(empty::<(), ()>());
-    unwrap!(res);
+    let mut core = unwrap!(Core::new());
```
povilasb: I intentionally wanted to keep the very first example (`connect`) simple - friendlier for newcomers.
Could we have to separate examples for this?

examples/connect.rs
```diff
@@ -78,60 +97,111 @@ impl fmt::Display for PeerId {
...
+                                .and_then(|line| {
+                                    let line = line.into_bytes();
+                                    peer_sink
+                                    .send((0, line))
+                                    .map(|peer_sink| Loop::Continue(peer_sink))
+                                    .map_err(|e| panic!("error sending message to peer: {}", e))
+                                })
+                            })
+                        };
+                        let reader = {
```
povilasb: That's a lot of nesting levels. Would you agree with me to extract some code to functions?

Cargo.toml
```diff
@@ -38,13 +38,14 @@ serde = "~1.0.10"
 serde_derive = "~1.0.10"
 serde_json = "~1.0.2"
 tiny-keccak = "~1.3"
-tokio-core = "~0.1.10"
+tokio-core = "=0.1.10"
```
povilasb: Hm, would be interesting to know why was this notation chosen instead of the default one?
You could put such changes in a separate commit and comment the reasons, this way I wouldn't ask in the future :)

src/net/protocol_agnostic/stream.rs
```diff
@@ -133,78 +136,101 @@ impl PaStream {
...
+                        UtpSocket::from_socket(udp_socket, &handle).map_err(
+                            UtpRendezvousConnectError::IntoUtpSocket,
+                        )?
+                    };
+                    trace!("returning utp socket.");
+                    Ok((utp_socket, utp_listener, addr))
                 })
         };
 
+
```
povilasb: redundant new line?

src/net/protocol_agnostic/stream.rs
```diff
@@ -133,78 +136,101 @@ impl PaStream {
...
-                .select2(connect)
-                .map_err(|either| match either {
-                    Either::A((err, _)) |
-                    Either::B((err, _)) => err,
-                })
-                .and_then(|either| match either {
-                    Either::A(((), connect)) => connect.into_boxed(),
-                    Either::B((stream, _)) => future::ok(stream).into_boxed(),
-                })
+            .while_driving(tcp_connect)
```
povilasb: This method is already large (~140 lines), do we really want to keep expanding it?

src/net/protocol_agnostic/stream.rs
```diff
@@ -133,78 +136,101 @@ impl PaStream {
...
+            .while_driving(udp_connect)
+            .map_err(|((e, _tcp_connect), _udp_connect)| e)
+            .and_then(move |((their_pk, tcp_connect), udp_connect)| {
+                if our_pk > their_pk {
+                    let utp_connect = {
+                        udp_connect
+                        .and_then(|(utp_socket, _utp_listener, addr)| {
+                            utp_socket
+                                .connect(&addr)
+                                //.map_err(UtpRendezvousConnectError::UtpConnect)
```
povilasb: I'd say let's not keep commented out code - it's cryptic for people who did not write it.

src/net/protocol_agnostic/stream.rs
```diff
@@ -133,78 +136,101 @@ impl PaStream {
...
+                    .map(|(stream, _buf)| stream)
+                    .into_boxed()
+                } else {
+                    let tcp_connect = tcp_connect.map(take_chosen);
+                    let utp_connect = {
+                        udp_connect
+                        .and_then(|(_utp_socket, utp_listener, addr)| {
+                            utp_listener
+                            .incoming()
+                            .filter(move |stream| stream.peer_addr() == addr)
```
povilasb: Does this filter connections to self? If so, how is this possible?

canndrew: No, it just makes sure we're accepting the connection from the address we hole-punched to.

povilasb: oh, good :)

src/net/peer/connect/mod.rs
canndrew: No, `config` is passed to a closure later on.

examples/connect.rs
canndrew: I've factored out the part which is irrelevant to making a connection into the `have_a_conversation` function. This also simplifies the example as a whole.

Cargo.toml
canndrew: That was an error and gets undone in a later commit.

src/net/protocol_agnostic/stream.rs
canndrew: It could be split into smaller functions. The thing is, if we're going to have functions which are only ever meant to be called from one place inside another function then it seems clearer to me to define those functions inside the larger function. eg.

```rust
fn foo() {
    fn implementation_detail_of_foo() {
        ...
    }
    ...
    let x = implementation_detail_of_foo();
    ...
}
```

Seems clearer than

```rust
fn foo() {
    ...
    let x = implementation_detail_of_foo();
    ...
}

fn implementation_detail_of_foo() {
    ...
}
```
 Because it gives `implementation_detail_of_foo` more precise scope. This is kind of what I was aiming for when splitting the implementation of that method into separate futures with

```rust
let pump_channels = {
    ...
};

let tcp_connect = {
    ...
};

let utp_connect = {
    ...
};
```

src/net/protocol_agnostic/stream.rs
canndrew: Another mistake that got fixed in a later commit. Sorry, it's just cumbersome to back-port small fixes when I've already layered other commits on top of the error. Unless you know of an easier way to do this?

canndrew: My workflow for that is

```
$ git checkout -b temp-branch
$ git checkout real-branch
$ git reset --hard commit-that-needs-to-be-changed
$ <make changes>
$ git commit --amend
$ git checkout temp-branch
$ git rebase --onto real-branch commit-that-needs-to-be-changed temp-branch
$ git checkout real-branch
$ git merge --ff-only temp-branch
$ git branch -d temp-branch
```

And then notice there was another spurious comment or something a few lines down and have to do it again.

src/net/peer/connect/mod.rs
povilasb: right, I got confused because `connect_directly()` call is in the same scope.
Turns out it's implementation passes config into closure. Wouldn't it be more readable, if  `connect_directly()` took reference, made a copy and passed that copy to a closure?

src/net/protocol_agnostic/stream.rs
povilasb: yeah, this kind of scoping sounds ok, but I think it hurts unit testing of those smaller functions.

src/net/protocol_agnostic/stream.rs
povilasb: wow, indeed that looks like a lot of work.
Personally, I'm using git interactive rebase, it's described in here: https://git-scm.com/book/id/v2/Git-Tools-Rewriting-History
In a nutshell, it works smth like this:
```
$ git checkout real-branch  # just explicitly saying where I am :)
$ git rebase -i HEAD~5  # opens editor that allows me to change 5 commits back from now
$ < I choose commit i want to edit and git checksout this commit>
$ <make changes>
$ git commit --amend
$ git rebase --continue
```
I could demonstrate that over hangout, if you'd like ;)

src/net/peer/connect/mod.rs
canndrew: `connect_directly` does take a reference, it's further down at the bottom of the function that `config` gets passed to a closure. Here's it needs to be owned so that the resulting future can be `'static`.

canndrew: See line 161

src/net/protocol_agnostic/stream.rs
canndrew: Thanks, I can copy what you've done there. I've never bothered to learn  `git rebase -i`, I guess I should because that's not the first nifty trick I've seen done with it.

examples/connect.rs
```diff
@@ -124,82 +124,40 @@ fn main() {
...
-                            Either::A((v, _next)) => void::unreachable(v),
-                            Either::B(((), _next)) => drop(service),
-                        })
-                        .map_err(|either| match either {
-                            Either::A((v, _next)) => v,
-                            Either::B((v, _next)) => v,
+                            service
+                                .connect(our_priv_info, their_pub_info)
+                                .map_err(|e| panic!("error connecting to peer: {}", e))
+                                .and_then(move |peer| have_a_conversation(service, peer))
```
povilasb: That's cool - really more readable, but I would still like to keep a minimal `connect` example and what if this one was a `chat` example instead?
Then we would have examples from least to most complex:
* connect - only connects two peers
* exchange_data - exchanges one message each
* chats - let's you send as many messages interactively as you want

povilasb:
Commit mesasges still don't comply to our conventions. Btw, it's easy to rewrite commit messages with git interactive rebase using `reword` function.

examples/chat.rs
```diff
@@ -0,0 +1,218 @@
...
+// found in the root directory of this project at LICENSE, COPYING and CONTRIBUTOR.
+//
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+//! This example demonstrates how to make a P2P connection using `crust`.
```
povilasb: Update docs for this example?

Issue #889, opened 2017-12-01
Don't call rendezvous_connect if we know we have an accessible listener.


canndrew:
Currently, crust will always call rendezvous_connect from the p2p crate when doing a rendezvous connect. However we don't need to do this if we know we have a publicly-accessible open port.

Some things to consider:
  * We need to make sure that p2p doesn't return false positives about when a port is publicly-accessible.
  * We need a way to make sure that the port on the router stays open.

povilasb:
This was implemented some time ago,

PR #919, opened 2018-01-17
Don't do rendezvous connects, if we have public IP

povilasb:
Addresses https://github.com/maidsafe/crust/issues/889
Basically, if there's a listener who successfully uses `bind_public()` from `p2p`, then `Crust` can be contacted directly and rendezvous connections should be avoided.

There's still one TODO though:
> // TODO(povilas): remove public address when listener gets dropped

but I figured this can be done with future PR.


maidsafe-highfive:

povilasb:
Also, note that how `Acceptor` is just a thin wrapper around `Listeners`: https://github.com/maidsafe/crust/pull/919/commits/5c10d642b1867dc1b448a84d48798845f5415cde#diff-5ed291e9fb97fc09364e19d6d2311f42R61

I'd like to remove `Acceptor` completely and rename `Listeners` to `Acceptor` instead. Then we'd have `Acceptor` and `Listener` structs instead of  `Acceptor`, `Listeners` and `Listener`.
@canndrew what do you think?

canndrew:
> I'd like to remove `Acceptor` completely and rename `Listeners` to `Acceptor` instead.

I guess that makes sense. There's quite a lot of indirection in there at the moment.

ustulation:
OS X hung, rest of CIs pass.

PR #920, opened 2018-01-18
React to dropped listeners

povilasb:
When listener is dropped it's public addresses are removed from list too.
This is necessary to properly determine if Crust is accessible trough direct connections.
This PR finalizes https://github.com/maidsafe/crust/pull/919 and also addresses https://github.com/maidsafe/crust/issues/889


ustulation:
OS X build hung. Rest of CIs pass

PR #922, opened 2018-01-25
Bootstrap order

### LINKED ISSUE #913: Use correct bootstrap order
When bootstrapping, make sure we try to connect to service-discovery peers first, then bootstrap cache peers, then hard-coded contacts. Also make sure that groups of peer addresses are shuffled.

canndrew:
Fix #913


canndrew:
Oops, wrong branch.

PR #924, opened 2018-02-01
Encryption

povilasb:
The very first steps into encryption. This PR definitely does not cover everything. It focuses on hole-punched connections, other connections should be intact and do not use encryption.

This PR introduces `CryptoContext` helper class to do encryption. Which has an optional encryption feature - you can use `CryptoContext::null()` to construct null encryptor which basically only serializes data without encryption.

Some future things to consider: bootstrap connection encryption, direct connection encryption and address query request/response encryption.


maidsafe-highfive:

canndrew:
All looks pretty straight-forward. I'm glad that we're finally making progress on this :)

PR #926, opened 2018-02-08
Encrypt bootstrap and direct connections

povilasb:
Basically what it says on the tin.
Also, worth mentioning that currently service discovery test is disabled, because it would be failing otherwise.
This will be fixed once encryption is implemented for service discovery.


maidsafe-highfive:

src/crypto.rs
```diff
@@ -43,58 +44,95 @@ quick_error! {
             description("Error decrypting message")
             display("Error decrypting message: {:?}", e)
         }
+        InvalidOperation(msg: String) {
```
canndrew: Can either make this take an enum or split it into two separate variants? I don't really like that it carries a string describing the error.

povilasb: sure thing, give me a minute, I'll amend that :)

canndrew (approved):

PR #928, opened 2018-02-13
MAID-2497: Fix p2p connections

povilasb:
After encryption changes there was some regression introduced: p2p connections wouldn't function anymore. The problem was that initial connect message was sent anonymously encrypted and then the socket would go to authenticated encryption. But it had to wait for anonymously encrypted connect message coming from remote peer first. Hence peers failed to decrypt any future connect messages and connections would fail.

This PR makes p2p connections use authenticated encryption right from the beginning and there's no such problem anymore.


maidsafe-highfive:

canndrew (approved):

PR #930, opened 2018-02-20
MAID-2500: echo address request encryption

povilasb:
Brings encryption to echo address requests/responses.


povilasb:
Currently this PR fails because it depends on p2p PR https://github.com/ustulation/p2p/pull/56 to get merged to master branch.

PR #932, opened 2018-02-26
chore/ci: update ci scripts to latest rust stable

m-cat:


maidsafe-highfive:
r? @canndrew

src/common/socket.rs
```diff
@@ -165,7 +165,7 @@ impl SockInner {
         }
 
         // the mio reading window is max at 64k (64 * 1024)
-        let mut buffer = [0; 65_536];
+        let mut buffer = [0; 1_0000];
```
canndrew: Why was this changed and shouldn't the comment be updated too?

canndrew:
Ugh, sorry for forgetting this. Am I right in thinking this is (almost) purely about updating dependencies?

src/common/socket.rs
canndrew: I gather it should be `0x1_0000`

canndrew: (Although I personally find `65_536` to be a lot more readable)

m-cat: @canndrew Exactly, I just forgot about the `0x` part :)

m-cat: > (Although I personally find 65_536 to be a lot more readable)

Yeah, this new clippy lint caught me by surprise -- but I do think it is more clear this way. `65_536` looks like a random number in decimal, but in hexadecimal its significance is obvious.

m-cat:
> Ugh, sorry for forgetting this. Am I right in thinking this is (almost) purely about updating dependencies?

Yes, in addition to upgrading Rust Stable.

src/common/socket.rs
canndrew: I dunno, 65536 looks pretty recognizable to me. Although 64 * 1024 would be even better still. But I'll merge this if this is what clippy wants.

m-cat: We can discuss disabling the lint if you don't think it's appropriate for the crate. Personally I agree with clippy here since I don't know the powers of 2 by heart. I fixed the typo and re-pushed the code, final call is up to you.

canndrew:
@m-cat I've approved it.

src/service_discovery/mod.rs
```diff
@@ -287,7 +289,7 @@ mod tests {
             unwrap!(
                 el1.send(CoreMessage::new(move |core, poll| {
                     unwrap!(
-                        ServiceDiscovery::start(core, poll, listeners_1, token_1, 65_530),
+                        ServiceDiscovery::start(core, poll, listeners_1, token_1, 0xFFFA),
```
ustulation: this is horrible - even if clippy is complaining i would suppress that warning in these places and leave them as is.

nbaksalyar: @m-cat I think we should disable this clippy lint after all.
E.g. this particular case is the port number and it makes not much sense to have it as a hexadecimal. So I'd agree with @canndrew and @ustulation about this lint not being very helpful at least for Crust.

src/common/socket.rs
```diff
@@ -165,7 +165,7 @@ impl SockInner {
         }
 
         // the mio reading window is max at 64k (64 * 1024)
-        let mut buffer = [0; 65_536];
+        let mut buffer = [0; 0x1_0000];
```
nbaksalyar: I agree with @canndrew, this one is better to be rewritten as just `64 * 1024`

Cargo.toml
```diff
@@ -7,29 +7,29 @@ license = "GPL-3.0"
 name = "crust"
 readme = "README.md"
 repository = "https://github.com/maidsafe/crust"
-version = "0.30.0"
+version = "0.31.0"
```
nbaksalyar: If the newer Rust compiler version doesn't require us to make any more changes apart from the ones required by Clippy, I'd say it's better to be left at 0.30.0 at this point. Same goes for dependencies if all of them are compatible with Rust 1.24.0 (i.e. if Crust builds fine on the latest stable). Otherwise we might need to check if the updated deps are compatible with the data that is stored on the current Alpha-2 network.

PR #942, opened 2018-03-20
feat/netsim: add netsim tests

canndrew:


src/tests/netsim_tests.rs
```diff
@@ -0,0 +1,389 @@
+
```
povilasb: We should probably add license header.

Cargo.toml
```diff
@@ -60,3 +60,11 @@ git = "https://github.com/ustulation/p2p"
 [dev-dependencies]
 docopt = "~0.8.3"
 hamcrest = "~0.1.5"
+
+[target.'cfg(target_os = "linux")'.dependencies]
+netsim = { version = "~0.1.6", optional = true }
+
+
```
povilasb: redundant new line?

src/tests/netsim_tests.rs
povilasb: Also, if file path is `src/tests/netsim_tests.rs`, do we have to say tests again? Sounds like redundant labeling for me, I'd go with `src/tests/netsim.rs` :)

src/tests/netsim_tests.rs
```diff
@@ -0,0 +1,389 @@
...
+                    })
+                    .while_driving(listening)
+                    .map_err(|(e, _)| e)
+                    .map(|((), _listening)| ())
+                })
+            }));
+            res.void_unwrap()
+        });
+
+        let rendezvous_server_node_1 = netsim::node::endpoint_v4(|ip| {
```
povilasb: Why do we need 2 rendezvous nodes?

src/tests/netsim_tests.rs
```diff
@@ -0,0 +1,389 @@
...
+                        .start_listening()
+                        .map_err(|e| panic!("error starting listeners: {}", e))
+                        .collect()
+                    };
+
+                    drop_rx_a0
+                    .and_then(|()| {
+                        drop_rx_b0
+                        .map(|()| {
+                            drop(service);
```
povilasb: Do we really have to manually drop Service? What if we just let it hanging until test finishes?

canndrew: We need to make sure the `Service` survives until the end of the test. Otherwise everything breaks.

src/tests/netsim_tests.rs
canndrew: Why not? I wouldn't have found that bug yesterday without two rendezvous nodes. It makes the test a bit more realistic.

povilasb: Hm, maybe, when I do some manual testing, I only have one rendezvous peer running, which seems to have no difference since all it does is return peer's address.
My motivation for having only 1 rendezvous peer is simplicity - less code :)

src/tests/netsim_tests.rs
povilasb: Oh, right :)

src/tests/netsim_tests.rs
canndrew: Yeah okay that's a good motivation :)

The purpose of these netsim tests though is to be as realistic as possible so that we'll know whether the code works in real life.

PR #949, opened 2018-04-12
Service discovery fixes

canndrew:
Send a UDP broadcast on each individual interface, rather than just to 255.255.255.255. This makes it work for me.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,6 +42,7 @@ extern crate config_file_handler;
 extern crate env_logger;
 extern crate future_utils;
 extern crate futures;
+extern crate get_if_addrs;
 #[cfg(test)]
 #[macro_use]
 extern crate hamcrest;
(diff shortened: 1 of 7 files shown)


povilasb:
Hmm, isn't that exactly what `255.255.255.255` is for?

canndrew:
I don't know what linux is doing with packets sent to 255.255.255.255, but they don't arrive anywhere. At least not for my machine. Each interface has a broadcast address you can send to though.

povilasb:
`255.255.255.255` is defined by https://tools.ietf.org/html/rfc919. But according to https://en.wikipedia.org/wiki/Broadcast_address broadcast addresses were slightly changed:
```
This method of using the all-ones address was first proposed by R. Gurwitz and R. Hinden in 1982.[2] The later introduction of subnets and Classless Inter-Domain Routing changed this slightly, so that the all-ones host address of each subnet is that subnet's broadcast address.[3]
```
Probably some routers comply with RFC 919 while others don't and that's why `255.255.255.255` works for me, but not for you.

PR #951, opened 2018-04-18
MAID-2628: filter loopback addresses

povilasb:
When two peers listen on the same port and 127.0.0.1 is included into peer connection info, remote peer will try to connect to itself. We don't want that.


povilasb:
We discussed with Spandan Sharma that loopback connections should be possible for the sole purpose of testing. Although, because loopback addresses should not "appear outside a host" (https://tools.ietf.org/html/rfc1122#page-30), will make a feature toggle for this in the future

Issue #966, opened 2018-06-02
Pass node public and secret keys to Service constructor


octavonce:
Currently, crust generates the node's keypair when building the Service object. This means that a node's public key will be generated each time the node is started and there is no way to pre-configure this.

A node failure would mean that the public key of the node changes making it's entry in every other node's bootstrap cache expire.

Also, why pass a UID to the service object when the public key is a good enough way to establish identity?

povilasb:
Yep, definitely makes sense. That's actually smth we recently discussed at MaidSafe and the work is gonna start soon. See https://maidsafe.atlassian.net/browse/MAID-2730 if interested :)

povilasb:
Also, see comments at: https://github.com/maidsafe/crust/pull/967

PR #967, opened 2018-06-02
Use public key as node uid

octavonce:
#966


maidsafe-highfive:

canndrew:
Hi! Thanks for this. Were actually in the process though of fixing #966 ourselves and part of this will mean keeping the generic `UID` parameter. As such I'm going to close this but you should be able to see our fix on the crust dev branch by the end of the week.

Thanks for the effort though :)!

canndrew:
It's still a while away from being production ready I'd say. If you want NAT traversal I'd suggest using the [`p2p` crate](http://github.com/ustulation/p2p)

`p2p` was split off from crust, it's purely for NAT traversal (it's what crust uses internally for NAT traversal) and it's a lot more mature than crust (it works today). Note that you need to use the one on github though as the crates.io version is ancient and pretty much an entirely different library.

PR #971, opened 2018-06-27
MAID-2794, MAID-2728, MAID-2799: safe crypto changes

canndrew:
For your consideration.

I've pointed Cargo.toml to git repos for now so that tests can build/pass.


src/net/protocol_agnostic/stream.rs
```diff
@@ -15,99 +15,105 @@
...
 }
 
-/// Protocol agnostic stream that yields length delimited frames of `BytesMut`.
-pub type FramedPaStream = Framed<PaStream, BytesMut>;
-
 #[derive(Debug)]
-pub enum PaStream {
-    Tcp(TcpStream),
-    Utp(UtpStream),
+pub enum PaStreamInner {
```
povilasb: Does it need to be public?

PR #972, opened 2018-07-02
MAID-2794: fix tests on windows

povilasb:


src/tests/compat_api.rs
```diff
@@ -245,247 +245,273 @@ fn start_two_services_exchange_data() {
...
+            &service1,
+            &event_rx0,
+            &event_rx1,
+            CrustUser::Client,
+        );
+
+        thread::sleep(Duration::from_secs(1));
+        let res = blacklisted_listener.accept();
+        match res {
+            Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => (),
```
ustulation: maybe interrupt error too ? They should probably go together most of the time. If you agree check in other places too pls

ustulation: @povilasb noted
> EINTR (`ErrorKind::Interrupted`) is retried automatically in Rust, see: https://github.com/rust-lang/rust/issues/11214

PR #973, opened 2018-07-03
MAID-2895: refactor for safe_crypto-aware routing

canndrew:
This removes the `UID` trait from the compat API. Also includes some general API cleanups.


povilasb:
Only windows tests are failing which I'm working on at the moment.

PR #975, opened 2018-07-13
MAID-2511: in-memory stream

povilasb:
This PR implements in-memory stream and `PaStream` (protocol agnostic stream) variant with it. That effectively allows us to easier test code depending on `PaStream`. That is also demonstrated with some code extracted and tested in compatibility layer.
Before the only way to construct `PaStream` was to actually make a real uTP or TCP connection and then use stream for testing. This is cumbersome, requires lots of boilerplate and is slower. Having in-memory stream avoids that with little overhead.

### DIFF
--- a/src/util/memstream.rs
+++ b/src/util/memstream.rs
@@ -1,0 +1,163 @@
+// Copyright 2018 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under the MIT license <LICENSE-MIT
+// https://opensource.org/licenses/MIT> or the Modified BSD license <LICENSE-BSD
+// https://opensource.org/licenses/BSD-3-Clause>, at your option. This file may not be copied,
+// modified, or distributed except according to those terms. Please review the Licences for the
+// specific language governing permissions and limitations relating to use of the SAFE Network
+// Software.
+
+//! Tokio compatible in-memory stream implementation. Helpful for testing.
+
+// TODO(povilas): move to future_utils crate - this module is generic and highly reusable.
+
+use priv_prelude::*;
+use std::cmp;
+use std::io::{self, Read, Write};
+
+/// In-memory stream implementing `AsyncWrite + AsyncRead`. This stream is not connected with
+/// any other stream. Meaning what you write to the stream you can later read from it.
+#[derive(Default, Debug)]
+pub struct EchoStream {
+    buf: Vec<u8>,
+    shutdown: bool,
+}
+
+impl Read for EchoStream {
+    /// Reads as much data as possible to the given buffer and removes that data from internal
+    /// buffer.
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        if self.buf.is_empty() {
+            if self.shutdown {
+                return Ok(0);
+            } else {
+                return Err(io::ErrorKind::WouldBlock.into());
+            }
+        }
+
+        let bytes_to_read = cmp::min(self.buf.len(), buf.len());
+        buf.iter_mut()
+            .zip(self.buf.drain(..bytes_to_read))
+            .for_each(|(dst, src)| *dst = src);
+        Ok(bytes_to_read)
+    }
+}
+
+impl Write for EchoStream {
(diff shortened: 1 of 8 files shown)


src/util/memstream.rs
```diff
@@ -0,0 +1,159 @@
...
+            use super::*;
+
+            mod when_stream_is_empty {
+                use super::*;
+
+                #[test]
+                fn when_stream_is_open_it_returns_would_block_error() {
+                    let mut stream = EchoStream::default();
+
+                    let mut data = [0; 4096];
```
ustulation: Maybe make it much lesser in size to not have a big mem footprint in test cases if possible - like here it does not really matter what size it is because we are asserting an `EWOULDBLOCK`
Similarly other places in the test code - nothing incorrect, just wasteful.

src/util/memstream.rs
```diff
@@ -0,0 +1,159 @@
...
+            } else {
+                return Err(io::ErrorKind::WouldBlock.into());
+            }
+        }
+
+        let mut bytes_read = 0;
+        while bytes_read < buf.len() && !self.buf.is_empty() {
+            buf[bytes_read] = self.buf.remove(0);
+            bytes_read += 1;
+        }
```
ustulation: Seems a little primitive being done this way and also not optimal - removing elt 1 by 1 from the beginning of the vector is not the most efficient thing to do, esp if you have information already about how many to remove which can make the impl do so optimisations. Something that came to mind 1st without much thought (so if you find something better then go for it) is the following:
```
let min = cmp::min(self.buf.len(), buf.len());
buf.iter_mut().zip(self.buf.drain(..min)).for_each(|(dst, src)| *dst = src);
```
should do it (and probably cleaner, safer and cripier) ?

Guess mem copy is even quicker but probably/unfortunately `unsafe` ? Anyway you can evaluate further.

src/util/memstream.rs
```diff
@@ -0,0 +1,159 @@
...
+impl AsyncWrite for EchoStream {
+    /// Analogous to sending FIN to myself.
+    /// This will unblock `read()` calls which will get 0 as result value.
+    fn shutdown(&mut self) -> io::Result<Async<()>> {
+        self.shutdown = true;
+        Ok(Async::Ready(()))
+    }
+}
+
+impl fmt::Debug for EchoStream {
```
ustulation: is it necessary ? the only extra field you are avoiding to print happens to be a bool. Does it make debugging a lot easier to avoid that forcing us to hand code a debug impl ?

src/util/test.rs
```diff
@@ -56,3 +59,14 @@ pub fn bootstrap_cache_tmp_file() -> OsString {
     path.push(file_name);
     path.into()
 }
+
+/// Constructs event sender/receiver pair.
+pub fn event_sender() -> (CrustEventSender, Receiver<Event>) {
```
ustulation: A method named `sender` returning a `receiver` is interesting :P .. maybe call it `event_channel()` or something else better if you agree.

ustulation (changes requested):

src/util/memstream.rs
povilasb: Crust required `PaStream` variants to impl `Debug`, not sure why, will check,

src/util/memstream.rs
povilasb: Yep, that was the least complex solution though, didn;t bother to optimize since that's for tests. If you think it's worth it, I'll do it :)

ustulation: Ah the quality of code shouldn't be different in production and in test - i would change it. Test is not a second class citizen when it comes to code quality, sure it can have asserts and other things not used in production OR some magic number (e.g., vec with 4 elements etc) , but not something that does not conform to idiomatic usage of Rust for e.g.

src/util/memstream.rs
ustulation: just to make it clear(er) just in case, i meant was manual impl is not needed - the default one should suffice. I'm not saying to get rid of `Debug` trait for `PaStream`.

src/util/test.rs
povilasb: mhm, sounds good to me

PR #976, opened 2018-07-13
MAID-2511: fix compat layer

povilasb:
Fixes `InactivityImeout` error handling in compatibility layer. This error is translated into`LostPeer` event.

### DIFF
--- a/src/compat/service.rs
+++ b/src/compat/service.rs
@@ -128,7 +128,7 @@ impl Service {
                         let handle1 = state.service.handle().clone();
                         let (drop_tx, drop_rx) = future_utils::drop_notify();
                         state.bootstrap_acceptor = Some(drop_tx);
-                        let acceptor = state.service.bootstrap_acceptor();
+                        let acceptor = state.bootstrap_acceptor();
                         let cm = state.cm.clone();
                         let event_tx = state.event_tx.clone();
                         handle0.spawn({
@@ -232,7 +232,6 @@ impl Service {
                 let f = {
                     let handle = state.service.handle().clone();
                     state
-                        .service
                         .bootstrap(blacklist, use_service_discovery, crust_user)
                         .map_err(|e| error!("bootstrap failed: {}", e))
                         .and_then(move |peer| {
@@ -453,6 +452,17 @@ impl Service {
             }));
         unwrap!(rx.recv())
     }
+
+    #[cfg(test)]
+    pub fn disable_peer_heartbeats(&mut self) {
+        let (done_tx, done_rx) = std::sync::mpsc::channel();
+        self.event_loop
+            .send(Box::new(move |state: &mut ServiceState| {
+                state.disable_peer_heartbeats = true;
+                unwrap!(done_tx.send(()));
+            }));
+        unwrap!(done_rx.recv());
+    }
 }
 
 pub struct ServiceState {
@@ -465,6 +475,9 @@ pub struct ServiceState {
     listeners: Option<DropNotify>,
     service_discovery: Option<ServiceDiscovery>,
     service_discovery_enabled: bool,
+
+    #[cfg(test)]
+    disable_peer_heartbeats: bool,
 }
 
 impl ServiceState {
@@ -479,6 +492,8 @@ impl ServiceState {
             listeners: None,
             service_discovery: None,
             service_discovery_enabled: false,
+            #[cfg(test)]
+            disable_peer_heartbeats: false,
         }
     }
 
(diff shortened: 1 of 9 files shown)


maidsafe-highfive:

src/tests/compat_api.rs
```diff
@@ -504,3 +505,50 @@ mod bootstrap {
...
+            config2,
+            sk2.clone()
+        ));
+        service2.disable_peer_heartbeats();
+
+        unwrap!(service2.start_bootstrap(HashSet::new(), CrustUser::Client));
+        expect_event!(event_rx2, Event::BootstrapConnect(_peer_id, _));
+        expect_event!(event_rx1, Event::BootstrapAccept(_peer_id, _));
+
+        thread::sleep(Duration::from_millis(5 * INACTIVITY_TIMEOUT_MS));
```
ustulation: Can we not just wait until we receive lost peer from the channel rather than compulsorily waiting for this duration ? Seems a little wasteful otherwise to just block the test for this duration - think the timeout is pretty low for tests but we shouldn't rely on it unless we are testing the event is not fired before the timeout or something but probably that's not the case here.
We can ofc use [rx_timeout(inactivity_timeout + 10ms or something)](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv_timeout) if we are concerned about hanging in case of a buggy production code.

src/compat/service.rs
```diff
@@ -222,19 +240,31 @@ impl Service {
...
                 let event_tx = state.event_tx.clone();
+                #[allow(unused_mut)]
                 let f = {
                     let handle = state.service.handle().clone();
                     state
                         .service
                         .bootstrap(blacklist, use_service_discovery, crust_user)
                         .map_err(|e| error!("bootstrap failed: {}", e))
+                        .map(move |mut peer| {
+                            #[cfg(test)]
```
ustulation: Not sure we can, but would see if we can avoid inlining selective compilation this way. Maybe we can structure it a little better to not have these sort of code at all ?

ustulation (changes requested):

src/tests/compat_api.rs
```diff
@@ -551,4 +590,36 @@ mod when_no_message_received_within_inactivity_period {
         };
         assert!(peer1_lost || peer2_lost);
     }
+
+    #[test]
+    fn when_heartbeats_exchanging_tcp_peer_stays_alive() {
+        let (event_rx1, event_rx2, _s1, _s2) =
+            bootstrap_and_do_nothing(tcp_addr!("0.0.0.0:0"), true);
+        thread::sleep(Duration::from_millis(5 * INACTIVITY_TIMEOUT_MS));
```
ustulation: Can we not just wait until we receive lost peer from the channel rather than compulsorily waiting for this duration ? Seems a little wasteful otherwise to just block the test for this duration - think the timeout is pretty low for tests but we shouldn't rely on it unless we are testing the event is not fired before the timeout or something but probably that's not the case here.
We can ofc use [rx_timeout(inactivity_timeout + 10ms or something)](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv_timeout) if we are concerned about hanging in case of a buggy production code.
Also check in other places if applicable.

ustulation (changes requested):

src/util/memstream.rs
```diff
@@ -0,0 +1,163 @@
...
+        self.shutdown = true;
+        Ok(Async::Ready(()))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    mod memstream {
```
ustulation: Is this necessary ? The tests are already under the module called `memstream` (i.e., this file)

src/tests/compat_api.rs
```diff
@@ -515,3 +505,113 @@ mod bootstrap {
...
+
+        (event_rx1, event_rx2, service1, service2)
+    }
+
+    #[test]
+    fn when_heartbeats_disabled_tcp_peer_emits_lost_peer_event() {
+        let (event_rx1, event_rx2, _s1, _s2) =
+            bootstrap_and_do_nothing(tcp_addr!("0.0.0.0:0"), false);
+
+        match event_rx1.recv() {
```
ustulation: If something's wrong in production code this is just going to hang - i think a better way would be to use rx-with-timeout and make it around +100ms of delta. Which is why the [previous review comment](https://github.com/maidsafe/crust/pull/976#discussion_r203076070) stated this explicitly.

Similarly in other applicable places (mainly below, but pls give it a scan).

src/tests/compat_api.rs
```diff
@@ -515,3 +505,113 @@ mod bootstrap {
...
+            res => panic!("unexpected event: {:?}", res),
+        };
+    }
+
+    #[test]
+    fn when_heartbeats_exchanging_tcp_peer_stays_alive() {
+        let (event_rx1, event_rx2, _s1, _s2) =
+            bootstrap_and_do_nothing(tcp_addr!("0.0.0.0:0"), true);
+
+        let timeout = Duration::from_millis(5 * INACTIVITY_TIMEOUT_MS);
```
ustulation: Cannot this be a few ms above the threshold - i.e., does it need to be 5 times ?

ustulation (changes requested):

src/util/memstream.rs
povilasb: oops, meant `mod echostream`. This nests test cases related only to this structure, so if there were other structures or global functions, they wouldn't go into this module. I'll rebase the #975.

PR #982, opened 2018-08-01
MAID-2506: fix/querier: call connect() on udp querier sockets

canndrew:


povilasb:
ah `cargo fmt` is failing :/

PR #990, opened 2018-08-20
MAID-3030: fix unwraps

povilasb:
Crust has a compatibility layer that runs tokio event loop with `crust::Service` in a separate thread and communicates with it via memory channel. If this thread dies for some reasons, any attempts to send commands to this compatibility layer thread will fail.
This PR makes sure we won't panic on failures but return an error and allow upper layers (Routing) to recover - create new Crust instance


maidsafe-highfive:

PR #991, opened 2018-08-21
[Maid-3107] chore/ci: update scripts to Rust stable 1.28.0 

m-cat:
nightly to 2018-07-07
clippy to 0.0.212
rustfmt to 0.99.2


src/nat/util.rs
```diff
@@ -38,13 +38,13 @@ pub fn enable_so_reuseport(_sock: &TcpBuilder) -> io::Result<()> {
 /// A replacement for `IpAddr::is_global` while we wait for that to enter stable.
 pub fn ip_addr_is_global(ip: &IpAddr) -> bool {
     match *ip {
-        IpAddr::V4(ref addr_v4) => ipv4_addr_is_global(addr_v4),
```
pierrechevalier83: I would probably put these in a separate commit (about making latest clippy happy) like the rustfmt change, but yymv

m-cat: Done!

PR #1005, opened 2018-10-02
Update p2p

povilasb:
Use the latest p2p crate version which doesn't attempt direct connections. `rendezvous_connect()` functions now always return public rendezvous address or fail.


povilasb:
Not applicable anymore: we're proceeding with mio implementation at this moment.

PR #1007, opened 2018-10-03
Fix connect_all

povilasb:
Make sure Peer is not dropped after connection attempt. Otherwise connections might be terminated immaturely. Also, during `connect_all()` Crust was not doing a proper connection handshake, which was expected by connection demultiplexer. This PR makes sure `connect_all()` executes the whole connection handshake process, otherwise demux will discard those connections cause it's waiting for HandhsakeMessage::Connect.


povilasb:
Not applicable anymore: we're proceeding with mio implementation at this moment.

PR #1070, opened 2018-11-22
Use Docker for Building on Travis

jacderida:
@S-Coyle has advised submitting this PR.

We have longer term plans to use Docker as a central part of our Travis and internal build processes, and Crust was a simple place we could prove the concept. There are several reasons for using Docker:

* An identical environment both internally and on the Travis side
* Minimises Travis files to a simple docker pull, docker run setup (only applies to Linux at this time)
* Gives us back more control by defining our environments inside the container
* Developers can use the Docker configuration to reproduce build and CI problems
* Reduced build times due to dependencies being cached in the containers

Points of interest:

* Dockerfile that defines an environment for building Crust. It bases from the [official Rust container](https://hub.docker.com/_/rust/) to minimise setup.
* A utility Makefile for building and running the container easily.
* The README has been updated to include instructions for building from source. It seemed kind of weird to provide some information for building with Docker but have no information for a 'normal' build, so I also added some info for that. Those instructions weren't intended to be a step by step guide, but rather some general helpful information.
* The multi line Bash scripts are refactored out into files. You can notice there are unfortunately several commits in this PR where I was trying to change the Travis configuration to get it to work with multi line scripts, but none of these combinations worked. It does make the Travis process slightly more convoluted to understand, but in any case, I actually prefer maintaining a script file rather than inline scripts in a YAML document.

If we were only running the build on Linux we could remove all the Rust setup from the Travis file, but we need to retain that for OSX. Sadly that adds cruft to the Linux build that isn't necessary, but having an OSX definition in the same file means we can't get around this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM rust:1.29.2
+
+RUN apt-get update -y && mkdir /target
+
+WORKDIR /usr/src/crust
+COPY . .
+RUN cargo build --release --verbose --target-dir /target
+
+CMD cargo test --release --verbose --target-dir /target
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,23 +17,8 @@ branches:
 cache:
   cargo: true
 before_script:
-  - if [ "${TRAVIS_OS_NAME}" == "osx" ]; then
-      set -x;
-      curl -sSL https://github.com/maidsafe/QA/raw/master/travis/cargo_install.sh > cargo_install.sh;
-      bash cargo_install.sh cargo-prune;
-    elif [ "${TRAVIS_OS_NAME}" == "linux" ]; then
-      set x;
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM rust:1.29.2
+
+RUN apt-get update -y && mkdir /target
+
+WORKDIR /usr/src/crust
+COPY . .
+RUN cargo build --release --verbose --target-dir /target
```
nbaksalyar: I think we can also pre-build the tests and it should save us a bit more CI time.
It should be sufficient to add just a single command:
```
RUN cargo build --tests --release --verbose --target-dir /target 
```

jacderida: Great, thanks Nikita! I'll do that just now.

PR #1080, opened 2018-12-05
resolves #837: update deprecated mio code

### LINKED ISSUE #837: Mio deprecated APIs
All deprecated miio apis seem too be dumped into [mio-more crate](https://github.com/carllerche/mio-more/tree/0fdf0024756acf12d204e2a98aea04fb7bbcde09) which looks very undocumented and not-maintained a.t.m. So track that to [remove this linter](https://github.com/maidsafe/crust/blob/ffe94e8849340c98f97729c4aa4c381cf3cc59e2/src/lib.rs#L41) and provide proper solution with latest mio.

cc @nbaksalyar @madadam

povilasb:
* Use mio_extras crate for channels and timer;
* don't use Ready::error() and Ready::hup() on non Unix systems;
* replace mio::tcp/udp with mio::net.


ustulation:
> don't use Ready::error() and Ready::hup() on non Unix systems;
more like don't use them at all. It's something that mio does extra and may not be portable. All errors should be handled by the corresponding trigger for read/write.

Cargo.toml
```diff
@@ -9,14 +9,21 @@ readme = "README.md"
 repository = "https://github.com/maidsafe/crust"
 version = "0.31.0"
 
+[[example]]
```
ustulation: doesn't this normally come at the end of the file ? Usually we are more interested to quickly know the deps, and those should be at the top.

src/common/mio_ext.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright 2018 MaidSafe.net limited.
```
ustulation: Not needed. only read and write readiness are portable.

Cargo.toml
povilasb: `cargo edit` reformatted it this way when I did `cargo add mio-extras`. Unless we don't want to use `cargo edit` anymore.

PR #1081, opened 2018-12-05
resolves #1072: remove hole punching code

### LINKED ISSUE #1072: Remove hole punching code
Currently Crust has some NAT traversal code. Hole punching related code will be replaced with p2p crate which has more robust implementation.

1. Remove any hole punching related code.
2. Retain IGD implementation.

povilasb:
A more robust hole punching implementation is inside p2p crate. It will be integrated into Crust in the future.


src/nat/mapped_tcp_socket/mod.rs
```diff
@@ -104,30 +99,13 @@ where
...
             igd_children,
-            stun_children: HashSet::with_capacity(mc.peer_stuns().len()),
             mapped_addrs,
             timeout: core.set_timeout(Duration::from_secs(TIMEOUT_SEC), CoreTimer::new(token, 0)),
             finish: Some(finish),
             phantom: PhantomData,
         }));
 
-        // Ask Stuns
-        for stun in mc.peer_stuns() {
```
ustulation: If you don't do this how will you get your public address to tell others where to connect ? (Orthogonal to NAT traversal, e.g., even in case of direct-connections how are you informing the peers about where they should connect to you ? The tests probably pass as they fallback to loopback due to `0.0.0.0` nature of the addresses just now ?)

povilasb: Right.. ok, I'll bring this back.

povilasb: ... for now, until p2p is not integrated yet :)

PR #1082, opened 2018-12-05
resolves #1075: service discovery improvements

### LINKED ISSUE #1075: Review service discovery code
We've witnessed occasional service discovery test failures on travis. Look for potential bugs and see if the code could be made more robust.

povilasb:
* Fixed UDP read to receive all packets until the socket blocks,
* added ability to run service discovery listener on a separate port.

### DIFF
--- a/src/service_discovery/mod.rs
+++ b/src/service_discovery/mod.rs
@@ -105,17 +105,32 @@ impl ServiceDiscovery {
     }
 
     fn read(&mut self, core: &mut Core, poll: &Poll) {
-        let (bytes_rxd, peer_addr) = match self.socket.recv_from(&mut self.read_buf) {
-            Ok((bytes_rxd, peer_addr)) => (bytes_rxd, peer_addr),
-            // TODO(povilas): handle WouldBlock as well
-            Err(ref e) if e.kind() == ErrorKind::Interrupted => return,
-            Err(e) => {
-                debug!("ServiceDiscovery error in read: {:?}", e);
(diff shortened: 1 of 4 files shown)


src/service_discovery/mod.rs
```diff
@@ -105,35 +111,40 @@ impl ServiceDiscovery {
...
+                    return;
+                }
+            };
 
-        match msg {
-            DiscoveryMsg::Request { guid } => {
-                if self.listen && self.guid != guid {
-                    self.reply_to.push_back(peer_addr);
-                    self.write(core, poll)
+            let msg: DiscoveryMsg = match deserialise(&self.read_buf[..bytes_rxd]) {
```
ustulation: maybe do this inline in the match arm of `recv_from` above.
This will ensure that
1. you don't discard all (from potentially various peers) but the last message
1. would blocks/interrupts after reading a valid message don't make you discard the messages read so far in this run of the loop

src/service_discovery/mod.rs
```diff
@@ -105,35 +111,40 @@ impl ServiceDiscovery {
...
+                Ok(msg) => msg,
+                Err(e) => {
+                    debug!("Bogus message serialisation error: {:?}", e);
+                    return;
```
ustulation: Will be a `continue` or NoOp (if there's nothing after the `recv_from()` match arm above, so it'll naturally continue the loop) once you have inlined it.

PR #1083, opened 2018-12-06
resolves #1073: integrate socket collection

### LINKED ISSUE #1073: Integrate socket-collection
Integrate [socket-collection](https://github.com/maidsafe/socket-collection) into Crust which will bring encryption as well.

povilasb:


src/main/connection_candidate.rs
```diff
@@ -103,7 +104,8 @@ impl<UID: Uid> ConnectionCandidate<UID> {
     fn done(&mut self, core: &mut Core, poll: &Poll) {
         let _ = core.remove_state(self.token);
         let token = self.token;
-        let socket = mem::replace(&mut self.socket, Socket::default());
+        let socket = mem::replace(&mut self.socket, Default::default());
+        let _ = poll.reregister(&socket, token, Ready::readable(), PollOpt::edge());
```
ustulation: hmm actually i want to get rid of this idiom of having to re-register to express exactly what we need etc. It's a laborious way to overly fine-tune and is prone to errors. The uniform idiom should be to just read in a loop. If you collect many msgs and only one is meant to be handled by the current state, then pass the remainder over to the next state. Similarly never re-register to write (or avoid it unless impossible to do without it). Just write instead. Start with registering once for read and write with edge notification and work your way from there.

This will however need a well thought out change so it's OK to do it in future PR. I'll put it as an issue.

src/service_discovery/mod.rs
```diff
@@ -112,15 +107,11 @@ impl ServiceDiscovery {
...
-                Err(ref e)
-                    if e.kind() == ErrorKind::Interrupted || e.kind() == ErrorKind::WouldBlock =>
-                {
-                    return
+            match self.socket.read_frm() {
+                Ok(Some((msg, peer_addr))) => {
+                    self.handle_incoming_msg(msg, peer_addr);
                 }
+                Ok(None) => return,
                 Err(e) => {
```
ustulation: hmm this is not correct anymore after your changes above. The deserialisation error due to 1 peer will now cause the collapse of entire ServDisc server and affect all other peers. So filter that separately and continue the loop in that case to cater to others.

src/service_discovery/mod.rs
```diff
@@ -130,26 +121,11 @@ impl ServiceDiscovery {
...
-            }
-        };
-
+    fn handle_incoming_msg(&mut self, msg: DiscoveryMsg, peer_addr: SocketAddr) {
         match msg {
             DiscoveryMsg::Request { guid } => {
                 if self.listen && self.guid != guid {
-                    self.reply_to.push_back(peer_addr);
-                    self.write(core, poll)
+                    self.respond_with_their_addr(peer_addr);
```
ustulation: Why not just use `self.write()` which takes an `Option` and pass it `Some()` ? Makes it uniform with code elsewhere in all our libs. Also a single place to read from and write to a socket per state is better as we need to only look there for anything. So bring the functionality of `respond_with..()` over here inline and remove that newly created function.

src/main/connection_candidate.rs
ustulation: [issue created here](https://github.com/maidsafe/crust/issues/1087) for future.

src/service_discovery/mod.rs
povilasb: yeah, you're right. I'll fix this.

PR #1084, opened 2018-12-10
WIP resolves #1078: implement encryption

### LINKED ISSUE #1078: Consolidate the peer representation in configs and cache
Previously to connect to a peer we merely required it's EP. Now with socket-collection and with the idea of nothing being plaintext on the wire we'll need more info to represent the peer. We should have the EP+PublicKey in the config and cache files.

povilasb:


src/main/connect/mod.rs
```diff
@@ -202,7 +185,7 @@ impl<UID: Uid> Connect<UID> {
     fn accept(&mut self, core: &mut Core, poll: &Poll) {
         loop {
             match unwrap!(self.listener.as_ref()).accept() {
-                Ok((socket, _)) => self.exchange_msg(core, poll, Socket::wrap(socket)),
+                Ok((socket, _)) => self.exchange_msg(core, poll, TcpSock::wrap(socket)),
```
ustulation: If you are not nat-traversing here, would you even have listeners to accept something ? I don't think this function should even be around any more.

Also do check and remove `self.listener` if no longer needed.

src/main/connection_candidate.rs
```diff
@@ -103,7 +104,8 @@ impl<UID: Uid> ConnectionCandidate<UID> {
     fn done(&mut self, core: &mut Core, poll: &Poll) {
         let _ = core.remove_state(self.token);
         let token = self.token;
-        let socket = mem::replace(&mut self.socket, Socket::default());
+        let socket = mem::replace(&mut self.socket, TcpSock::default());
```
ustulation: `Default::default()`

src/main/connection_listener/exchange_msg.rs
```diff
@@ -372,7 +373,8 @@ impl<UID: Uid> ExchangeMsg<UID> {
             }
         }
 
-        match self.socket.write(poll, self.token, msg) {
+        // TODO(povilas): use authed encryption
```
ustulation: This should always be authenticated encryption as it's _responding_ to someone meaning there was a prior read where that someone must have given their keys by now.

So all of [these reads](https://github.com/maidsafe/crust/pull/1084/files#diff-5f3070043b7a5b205deab9f8f768a619R130) should also give the public-key to this state.

src/main/connection_listener/exchange_msg.rs
```diff
@@ -391,7 +393,7 @@ impl<UID: Uid> ExchangeMsg<UID> {
 
         match self.next_state {
             NextState::ActiveConnection(their_uid, peer_kind) => {
-                let socket = mem::replace(&mut self.socket, Socket::default());
+                let socket = mem::replace(&mut self.socket, TcpSock::default());
```
ustulation: `Default::default()`

src/main/connection_listener/exchange_msg.rs
```diff
@@ -426,7 +428,7 @@ impl<UID: Uid> ExchangeMsg<UID> {
                     }
                 };
 
-                let socket = mem::replace(&mut self.socket, Socket::default());
+                let socket = mem::replace(&mut self.socket, TcpSock::default());
```
ustulation: `Default::default()`

src/main/connection_listener/mod.rs
```diff
@@ -146,11 +162,19 @@ impl<UID: Uid> ConnectionListener<UID> {
         loop {
             match self.listener.accept() {
                 Ok((socket, _)) => {
+                    let mut socket = TcpSock::wrap(socket);
+                    if let Err(e) = socket.set_decrypt_ctx(DecryptContext::anonymous_decrypt(
```
ustulation: This might be tricky. You cannot blanket set this as it depends on what the message is. If someone is directly connecting to us they would use authenticated encryption as crust allows _active_ connections only between known peers. If it's a bootstrap accept it will be anonymous decryption as we don't know the enc_pk of bootstrapper until the handshake.

Maybe there should be a mode in sock-coll to say try-all and report which one was successful. Then in exchg_msg here, enforce that if anon was successful the msg must be Bootstrap-req or IpEcho-req. For a connect-attempt it must be authenticated encryption.

src/lib.rs
```diff
@@ -70,6 +70,7 @@
         too_many_arguments
     )
 )]
+#![recursion_limit = "128"]
```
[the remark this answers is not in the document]
povilasb: yep, didn't really investigate why it actually requires that. Any ideas?

src/main/connection_listener/mod.rs
povilasb: Actually, the direct connections will always start with anonymous encryption and then will upgrade to authenticated one. That's because the listener is not aware of who will be connecting: anybody can connect to the same listener. Given that, we should be good in this case with no extra changes/hacks in socket-collection :)

src/service_discovery/mod.rs
```diff
@@ -95,45 +104,38 @@ impl ServiceDiscovery {
...
-
-        match msg {
-            DiscoveryMsg::Request { guid } => {
-                if self.listen && self.guid != guid {
-                    self.reply_to.push_back(peer_addr);
-                    self.write(core, poll)
+        loop {
+            let (msg, peer_addr) = match self.socket.read_frm() {
+                Ok(Some((msg, peer_addr))) => (msg, peer_addr),
+                Ok(None) => return,
```
ustulation: Isn't this a wouldblock scenario and you should break if you have collected something in previous run of the loop (if collecting in a vector for e.g.) ?

src/service_discovery/mod.rs
```diff
@@ -95,45 +104,38 @@ impl ServiceDiscovery {
...
+                    debug!("ServiceDiscovery error in read: {:?}", e);
+                    self.terminate(core, poll);
+                    return;
+                }
+            };
+
+            match msg {
+                DiscoveryMsg::Request { our_pk: their_pk } => {
+                    if self.listen && self.our_pk != their_pk {
+                        self.reply_to.push_back(peer_addr);
```
ustulation: Don't think you need to do all this book keeping if using `UdpSock` from sock-coll.

src/service_discovery/mod.rs
```diff
@@ -149,17 +151,11 @@ impl ServiceDiscovery {
...
+            match self.socket.write_to(Some((resp, peer_addr, 0))) {
                 // UDP is all or none so if anything is written we consider it written
-                Ok(_bytes_send) => (),
-                Err(ref e)
-                    if e.kind() == ErrorKind::Interrupted || e.kind() == ErrorKind::WouldBlock =>
-                {
-                    self.reply_to.push_front(peer_addr)
-                }
+                Ok(true) => (),
+                Ok(false) => self.reply_to.push_front(peer_addr),
```
ustulation: Do you need to do this book-keeping ? `UdpSock` should have internally buffered it for a resent when the readiness for socket fires again.

src/main/connection_listener/mod.rs
ustulation: In order to initiate an _active_ connection attempt, you are required to know the peer completely. So it will give you authenticated encrypted msg. Only bootstrapping nodes will give you anon as you as a proxy are not expected to know the pks of all the future bootstrappers. Same with IP-echo service.

src/lib.rs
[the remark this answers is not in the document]
ustulation: Na, it's to prevent recursion in macro. Should be fine.

src/main/connection_listener/mod.rs
ustulation: Hmm.. So
1. you _connect_ to someone who you know and who knows you
1. you bootstrap or ip-echo-req someone who you know but who doesn't know you

However `1` could lead to some rejections in connections esp if the other peer hasn't yet realised it needs you but you have. Eventually it will be connected because when it realises it will connect to you who has already realised it. This will need the upper layer to tell crust from whom it can _accept_ a direct connection request which is currently not the behaviour in Routing.

Let's actually leave this just now and discuss the intended behaviour.

ustulation: So to be clear: for this change-set go with what you were doing - leave the overall behaviour unchanged.

src/service_discovery/mod.rs
```diff
@@ -29,48 +28,58 @@ use std::u16;
...
 
 pub struct ServiceDiscovery {
     token: Token,
-    socket: UdpSocket,
+    socket: UdpSock,
     remote_addr: SocketAddr,
     listen: bool,
-    read_buf: [u8; 1024],
-    our_listeners: Arc<Mutex<Vec<SocketAddr>>>,
-    seek_peers_req: Vec<u8>,
```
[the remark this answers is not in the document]
povilasb: Makes sense, but we can't really do this with `UdpSock` from socket-collection since it does message serialization itself.

src/service_discovery/mod.rs
povilasb: meaning just write directly to `UdpSock`, because it will buffer writes anyway? :)

src/service_discovery/mod.rs
```diff
@@ -95,45 +104,38 @@ impl ServiceDiscovery {
...
-        };
-
-        match msg {
-            DiscoveryMsg::Request { guid } => {
-                if self.listen && self.guid != guid {
-                    self.reply_to.push_back(peer_addr);
-                    self.write(core, poll)
+        loop {
+            let (msg, peer_addr) = match self.socket.read_frm() {
+                Ok(Some((msg, peer_addr))) => (msg, peer_addr),
```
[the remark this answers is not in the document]
povilasb: the processing indeed is inline: once the message is received, the peer address is buffered for future responses when socket becomes writable. See: https://github.com/maidsafe/crust/pull/1082/commits/2cd2749a9ae6944818ea550399f7fb29f5084069#diff-1ea1325c533fb71838b740a400352538R123

src/service_discovery/mod.rs
povilasb: mhm and that's what it exactly does: breaks the loop and returns from the function.

src/service_discovery/mod.rs
ustulation: It'll buffer it only if it can't write it. So `write()` here should take an option of message. This will pass `Some()` for that and write being called from the readiness handler will pass `None` meaning if it's a spurious wake then sock-coll will internally do nothing and if it was due to a prior WB then sock-coll would internally re-attempt to write the previously buffered stuff. [Check here](https://github.com/maidsafe/socket-collection/blob/e1ba943ded9b3e3b29bc14ba12b517aff9729b7b/src/udp.rs#L480).

So follow other modules [like here](https://github.com/maidsafe/crust/blob/a3d0deeeaf47739b482790e6074476af965b8162/src/main/active_connection.rs#L153) and a `None` case for it [here](https://github.com/maidsafe/crust/blob/a3d0deeeaf47739b482790e6074476af965b8162/src/main/active_connection.rs#L178)

povilasb:
Closed in favor of https://github.com/maidsafe/crust/pull/1091 . I didn't want to loose all these comments, so I opened a new PR that squashed all encryption related commits which were meant to be a single commit, but turned out to be multiple commits to embrace the iterative development :)

PR #1091, opened 2018-12-13
resolves #1078: encrypt everything on the wire

### LINKED ISSUE #1078: Consolidate the peer representation in configs and cache

povilasb:
* Uses IP address  + public key pair to identify remote peers,
* encrypts all requests (except service discovery)


src/main/connection_listener/mod.rs
```diff
@@ -147,17 +166,27 @@ impl<UID: Uid> ConnectionListener<UID> {
         loop {
             match self.listener.accept() {
                 Ok((socket, _)) => {
+                    let mut socket = TcpSock::wrap(socket);
+                    if let Err(e) = socket.set_decrypt_ctx(DecryptContext::anonymous_decrypt(
+                        self.our_pk,
+                        self.our_sk.clone(),
+                    )) {
+                        error!("Failed to set decryption context: {}", e);
```
ustulation: unless you are terminating the app, it should be `warn!` for logic-errors, unexpected-results-which-cannot-be-easily-explained-and-might-cause-UB-now (as is the case here) etc.

src/main/connect/exchange_msg.rs
```diff
@@ -90,15 +95,22 @@ impl<UID: Uid> ExchangeMsg<UID> {
...
                 if their_uid != self.expected_id || name_hash != self.expected_nh {
                     return self.handle_error(core, poll);
                 }
                 let _ = core.remove_state(self.token);
                 let token = self.token;
-                let socket = mem::replace(&mut self.socket, Default::default());
 
-                (*self.finish)(core, poll, token, Some(socket));
+                let mut socket = mem::replace(&mut self.socket, Default::default());
+                if socket
```
ustulation: It might be better to use `match` in these cases to log errors (in `debug!()`) to help debuggability in case something goes wrong.

src/main/bootstrap/try_peer.rs
```diff
@@ -80,9 +89,17 @@ impl<UID: Uid> TryPeer<UID> {
             Ok(Some(Message::BootstrapGranted(peer_uid))) => {
                 let _ = core.remove_state(self.token);
                 let token = self.token;
-                let socket = mem::replace(&mut self.socket, Default::default());
-                let data = (socket, self.peer, peer_uid);
-                (*self.finish)(core, poll, token, Ok(data));
+
+                let mut socket = mem::replace(&mut self.socket, Default::default());
+                if socket
```
ustulation: It might be better to use match in these cases to log errors (in debug!()) to help debuggability in case something goes wrong.

src/main/connection_listener/mod.rs
povilasb: mmk, where are have you read those definitions from though? :)

ustulation: Ah it's the general convention we are following really.

ustulation: Actually i did a quick search and [landed on this](https://stackoverflow.com/a/2031209/1060004). It sort of matches my take on the log levels, if you shift by 1, meaning that since the link has an extra level `FATAL` which we don't if you consider our-error=their-fatal, our-warn=their-error and so on (shift by 1) it seems pretty close to the suggestion here.

PR #1093, opened 2018-12-21
resolves #1074: use bootstrap cache

### LINKED ISSUE #1074: Use and prune bootstrap cache
1. Backport bootstrap cache implementation from Crust `dev` branch.
2. Search for peers during bootstrap in such order:
  2.1. service discovery
  2.2. cached peers
  2.3. hard coded contacts
3. Shuffle cached peers and hard coded contacts individually
4. Prune the ones in cache which fail connection
5. Don't add peers already in hard-coded-contacts to bootstrap cache

povilasb:
Most of the bootstrap cache related code was implemented by https://github.com/maidsafe/crust/commit/9a4c322a7b1be0ccb9da7698618b67bb3fae3ef2
This PR makes sure the correct use of bootstrap cache which resolves #1074


src/main/bootstrap/mod.rs
```diff
@@ -160,14 +161,7 @@ impl<UID: Uid> Bootstrap<UID> {
...
             Ok((socket, peer_info, peer_id)) => {
-                {
-                    let bootstrap_cache = core.user_data_mut();
-                    bootstrap_cache.put(peer_info);
-                    if let Err(e) = bootstrap_cache.commit() {
-                        info!("Failed to write bootstrap cache to disk: {}", e);
-                    }
-                }
-
+                cache_peer_info(core, peer_info, &self.config);
```
ustulation: Why do we want to cache folks here ? If they are in HCC, we don't want to cache them. If they are due to ServDisc, we don't want to cache them (likely LAN IP anyway and temporary too, plus ServDisc by itself is fine in (re)finding them). If they are from bootstrap cache, well then they are already in the cache.

Rather what we should be doing is to remove the ones who fail to be contacted which i can see is already implemented in the failure arm, so that should be sufficient ?

PR #1095, opened 2018-12-27
Upgrade to rust 2018 edition on dev branch

pierrechevalier83:
* Upgrade compiler
* Switch to 2018 edition
* Fix compilation, fmt and clippy warnings
* Remove nightly from CI


povilasb:
PR looks good to me. Although, we are not using the `dev` branch anymore, I'd still say merge it, if we decide we want to at some point :)

PR #1102, opened 2019-01-04
resolves #1086: configurable external reachability test

### LINKED ISSUE #1086: Configurable external reachability test
Currently Crust will test for external peer reachability during bootstrap. This test includes a physical TCP connection attempt to the peer that is trying to bootstrap. 

We would like to make several changes to the existing semantics to make external reachability test more robust:

1. Right now the client itself asks to test it's external reachability. Make connection listener test external reachability for [Node](https://github.com/maidsafe/crust/blob/404864436940f428567105e38549399bf3f6a864/src/common/mod.rs#L32) peers instead.
2. Never test external reachability for peers with [Client](https://github.com/maidsafe/crust/blob/404864436940f428567105e38549399bf3f6a864/src/common/mod.rs#L35) role.
3. Expose function in `Service` to enable/disable external reachability checks in connection listener.
4. By default external reachability test is always executed for `Node` peers.
5. In addition external reachability should also be tested during `Service::connect()`.
6. Currently external reachability test is merely a TCP connectivity test. Update it to include encrypted handshake to make sure, the connection is a valid crust connection.

povilasb:
1. External reachability test can now be executed during regular connections.
2. This test is optional and configured via parameter of `Service::connect()/start_bootstrap()` methods.

### DIFF
--- a/examples/crust_peer.rs
+++ b/examples/crust_peer.rs
@@ -501,7 +501,7 @@ fn main() {
                             continue;
                         }
                     };
-                    unwrap!(unwrap!(service.lock()).connect(our_info, their_info));
+                    unwrap!(unwrap!(service.lock()).connect(our_info, their_info, false));
                 }
                 UserCommand::Send(peer_index, message) => {
                     let network = unwrap!(network.lock());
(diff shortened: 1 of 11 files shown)


povilasb:
We had some discussions internally and this PR doesn't reflect our goals, so closing it. See latest PR https://github.com/maidsafe/crust/pull/1104

PR #1104, opened 2019-01-08
resolves #1086: configurable external reachability test

### LINKED ISSUE #1086: Configurable external reachability test

povilasb:


src/main/connect/mod.rs
```diff
@@ -41,6 +42,7 @@ pub struct Connect<UID: Uid> {
     event_tx: crate::CrustEventSender<UID>,
     our_pk: PublicEncryptKey,
     config: CrustConfig,
+    our_addrs: HashSet<SocketAddr>,
```
ustulation: call these vars `our_direct_listeners` pls to make it less confusing to what it is.

src/main/connection_listener/exchange_msg.rs
```diff
@@ -376,6 +422,61 @@ impl<UID: Uid> ExchangeMsg<UID> {
...
+                Ok(peer_addr) => addr.ip() == peer_addr.ip(),
+                Err(e) => {
+                    debug!(
+                        "Could not obtain IP Address of peer: {:?}. Denying handshake.",
+                        e
+                    );
+                    false
+                }
+            })
+            .collect()
```
ustulation: 1. If someone sends huge number of direct-listeners, it's going to make us waste a lot of resource trying out connecting. We should limit these to a reasonable number - say 2 or 3 global IPs.
2. There is no use sending local IPs for ext-reachability. So prune the list before hand instead of doing it everytime in connects (see review there). Here just discard the connection attempt if you find the node gives you local IPs.
3. Nodes from the same LAN will not be able to connect to each other if ext-reachability is on. This i think is overkill. So if you detect that all the octates are same as your ext addr, then bypass the ext-reachability test on that IP. If this is the only IP presented or all IPs match in octates with all your IPs then bypass the ext-reachability all-together. See if you find any weakness that is exploitable due to this though, before you implement it.

src/main/connect/exchange_msg.rs
```diff
@@ -46,6 +48,7 @@ impl<UID: Uid> ExchangeMsg<UID> {
         cm: ConnectionMap<UID>,
         our_pk: PublicEncryptKey,
         shared_key: SharedSecretKey,
+        our_addrs: HashSet<SocketAddr>,
```
ustulation: There is no use sending local IPs for ext-reachability. So prune the list before hand.

Also call these `our_global_direct_listeners` as noted elsewhere too.

src/main/config_handler.rs
```diff
@@ -58,6 +58,8 @@ pub struct Config {
 /// Developer options
 #[derive(PartialEq, Eq, Debug, Serialize, Deserialize, Clone, Default)]
 pub struct DevConfig {
+    // TODO(povilas): remove this field since there is a formal way to disable ext reachability
```
ustulation: Yes this is good to be removed now.

src/common/message.rs
```diff
@@ -7,19 +7,36 @@
...
+    EchoAddrResp(SocketAddr),
     ChooseConnection,
-    Connect(UID, NameHash, PublicEncryptKey),
+    /// First connect handshake message which goes both ways: peer that initiates connection
+    /// sends this message to remote peer, which then responds with the same message, but its own
+    /// information.
+    /// When `Connect` is used as response message address list is ignored and can be empty. When
+    /// it is a request message, it should contain all the addresses we are listening on. Those
+    /// addresses will be used for external reachability test.
+    // TODO(povilas): consider splitting Connect into ConnectRequest and ConnectResponse
```
ustulation: If you don't do this, see if this can be exploited by even coding errors. So both sides keep sending a non-empty list and they go into a loop of trying ext-reachability test of each other.

povilasb: Connection will function properly. The reason I made it notable with a todo note is that it would be readable. Currently we use `Message::Connect` when peer A connects to peer B, and then when peer B responds to peer A.  the Connect message that acts as a response will be sent to the socket connecting A ---> B. So it won't be handled by connection listener and ext reachability test won't be executed. The problem is solely readability. It is definitely easier to reason about the code when we have explicit request/response messages.
Anyway, I could have done this right away, but I feel like this change is better done as a separate refactoring PR. Otherwise, this PR will get fat :d

src/main/connection_listener/exchange_msg.rs
povilasb: 3. Exactly, but isn't this what external reachability test is for? :) Or should we skip external reachability for `Node` peers connecting on LAN? I don't think this such behavior will be used in production and we have the possibility to disable external reachability tests during testing anyway.

src/main/bootstrap/try_peer.rs
```diff
@@ -69,7 +70,13 @@ impl<UID: Uid> TryPeer<UID> {
             peer,
             socket,
             request: Some((
-                Message::BootstrapRequest(our_uid, name_hash, ext_reachability, our_pk),
+                Message::BootstrapRequest(
+                    our_uid,
+                    name_hash,
+                    our_global_direct_listeners,
+                    our_role,
```
ustulation: hmm what's the use of supplying our addresses if our role is that of a client ? Wasn't the enum you removed, better ?

src/main/connect/exchange_msg.rs
```diff
@@ -46,6 +48,7 @@ impl<UID: Uid> ExchangeMsg<UID> {
         cm: ConnectionMap<UID>,
         our_pk: PublicEncryptKey,
         shared_key: SharedSecretKey,
+        our_gloval_direct_listeners: HashSet<SocketAddr>,
```
ustulation: typo `global`

src/main/connection_listener/exchange_msg.rs
```diff
@@ -333,10 +325,62 @@ impl<UID: Uid> ExchangeMsg<UID> {
...
+        child: Token,
+        res: Result<UID, ()>,
+    ) {
+        let _ = self.reachability_children.remove(&child);
+        if let Ok(their_uid) = res {
+            self.terminate_childern(core, poll);
+            return self.send_connect_grant(core, poll, their_uid);
+        }
+        if self.reachability_children.is_empty() {
+            trace!("External reachability test failed, terminating connection.");
```
ustulation: these should probably be debugs now as it's become more important to note these failures.

src/main/connection_listener/mod.rs
```diff
@@ -31,6 +30,8 @@ use std::sync::{Arc, Mutex};
 const LISTENER_BACKLOG: i32 = 100;
 
 /// Accepts connections and transitions each connection into `ExchangeMsg` state.
+/// Optionally will make `ExchangeMsg` to test for peer external reachability. This behavior
+/// is disabled by default.
```
ustulation: But it seems you have enabled it by default [below at line 168 in this file in this diff](https://github.com/maidsafe/crust/pull/1104/files#diff-99e3273d6f63d6af85f9577cc6dcfc54R168)

src/main/bootstrap/try_peer.rs
povilasb: it makes sure the decision is made by listener and that listener always have data required to do so, if we decide to change it's behavior. Anyway, if that seems very unlikely, I could simply make global addresses an optional variable.

Also, for the reference 
```
enum ExternalReachability {
    Required, NotRequired
}
```
doesn't sound OK to me, cause now we are not asking for ext reachability test, rather we're giving the necessary info to do so.

src/main/connection_listener/mod.rs
povilasb: oops, good catch :+1: Just to clarify, it's intended to be enabled by default, so I will update the docs :)

src/common/mod.rs
```diff
@@ -35,6 +36,28 @@ pub enum CrustUser {
...
+    /// public endpoints.
+    Node(HashSet<SocketAddr>),
+    /// `Client` peers don't include any addresses, because they are never tested for external
+    /// reachability.
+    Client,
+}
+
+impl BootstrapperRole {
+    /// Converts into `CrustUser`.
+    pub fn as_crust_role(&self) -> CrustUser {
```
ustulation: Insn't `Into` (here) and `From` (for `CrustUser` for symmetry) more natural for these sort of things ?

PR #1105, opened 2019-01-08
Change Boostrap<Uid> peers' field type from Vec<PeerInfo to HashSet<P…

### LINKED ISSUE #1090: Possibility of multiple localhost address during bootstrap
Since ServDisc returns loopback address too if a listener is bound to it, multiple machines in a LAN might respond to a seeking/beaconing peer and they might all have loopbacks in their response. So in [places like these](https://github.com/maidsafe/crust/blob/bafde11bd33474ca3cfe2357d9b2d41364cdb831/src/main/bootstrap/mod.rs#L238) one might end up with multiple loopback addresses and try all of them (where at-most one will succeed if at all) which is wasteful. Either filter to remove duplicates there or turn [this vec](https://github.com/maidsafe/crust/blob/bafde11bd33474ca3cfe2357d9b2d41364cdb831/src/main/bootstrap/mod.rs#L42) into a `HashSet`.

As usual, analyse this does not break anything else.

1uka:
I decided that it would be better to use `HashSet<PeerInfo>` instead of going through a vector and filtering to remove duplicates.

Should resolve #1090 .

### DIFF
--- a/src/main/bootstrap/mod.rs
+++ b/src/main/bootstrap/mod.rs
@@ -47,7 +47,7 @@ const MAX_CONTACTS_EXPECTED: usize = 1500;
 pub struct Bootstrap<UID: Uid> {
     token: Token,
     cm: ConnectionMap<UID>,
-    peers: Vec<PeerInfo>,
+    peers: HashSet<PeerInfo>,
     name_hash: NameHash,
     ext_reachability: ExternalReachability,
     our_uid: UID,
@@ -118,7 +118,7 @@ impl<UID: Uid> Bootstrap<UID> {
     }
 
     fn begin_bootstrap(&mut self, core: &mut EventLoopCore, poll: &Poll) {
-        let peers = mem::replace(&mut self.peers, Vec::new());
+        let peers = mem::replace(&mut self.peers, HashSet::new());
         if peers.is_empty() {
             let _ = self.event_tx.send(Event::BootstrapFailed);
             return self.terminate(core, poll);
@@ -316,8 +316,8 @@ fn shuffled_bootstrap_peers(
     cached_peers: HashSet<PeerInfo>,
     config: CrustConfig,
     blacklist: HashSet<SocketAddr>,
-) -> Vec<PeerInfo> {
-    let mut peers = Vec::with_capacity(MAX_CONTACTS_EXPECTED);
+) -> HashSet<PeerInfo> {
+    let mut peers = HashSet::with_capacity(MAX_CONTACTS_EXPECTED);
     let mut rng = rand::thread_rng();
 
     let mut cached: Vec<_> = cached_peers.iter().cloned().collect();


ustulation:
@1uka Hey thanks of the PR ! Unfortunately it won't do the right thing.
1. The sources of that vec of peer-info are sets (denoting unique contents) so the vec is very likely going to be unique anyway.
2. However the main reason it'll be wrong is if you go through the code to see how it's manipulated before it's returned by the function whose signature you changed you'll notice that it does specific shuffling. Putting it into a set will interfere with that by (re)ordering it either deterministically (BTree) or unspecific-ally (Hash), which we don't want. So places where the existing ordering needs to be preserved exp. in places such as these, it's far easier and better to stick to a vector (else you need some arcane/verbose ord-trait implementations etc.)

1uka:
Oh, sorry I guess I hurried with this PR. Would you like me to try and solve it by filtering the listeners [here](https://github.com/maidsafe/crust/blob/bafde11bd33474ca3cfe2357d9b2d41364cdb831/src/main/bootstrap/mod.rs#L238), as you have mentioned in the issue description?

1uka:
So, just to make sure that I understand before I make the changes:

[This](https://github.com/maidsafe/crust/blob/master/src/service_discovery/mod.rs#L152) is the place where `ServiceDiscovery` sends the addresses, and [this](https://github.com/maidsafe/crust/blob/master/src/main/bootstrap/mod.rs#L246) is the only place where they need to be filtered, right?

And instead of filtering for duplicates, I should ignore loopback addresses (no matter what the port) ?

There are 2 ways I can see this being done:

```rust
        while let Ok(listeners) = rx.try_recv() {
            let peers: Vec<PeerInfo> = listeners
                .into_iter()
                .filter(|listener| !listener.addr.ip().is_loopback())
                .collect();
            self.peers.extend(peers)
        }
```
And if we want to also filter for duplicates, we can change the filter condition like so:
```rust
filter(|listener| !(listener.addr.ip().is_loopback() || self.peers.contains(listener)))
```
But, the call to `contains()` is O(n) and it will be expensive if we receive a lot of listeners.

Considering the scenario where there are multiple Crust instances and you want to store multiple loopback addresses, the condition can be this:
```rust
!(listener.addr.ip().is_loopback() && self.peers.contains(listener))
```
But if we receive a loopback address from a different machine via `ServiceDiscovery`, the problem is still not solved.

Which of the above propositions do you see most fitting?

povilasb:
I'd say let's not completely filter out loopback addresses - this way we couldn't test code on localhost. That could be unexpected in some cases.

How about receiving all those addresses into a `HashSet` first and the append to all peers list:
```
let discovered_peers: HashSet<PeerInfo> = rx.iter().flatten().collect();
self.peers.extend(discovered_peers);
```
On average insert into `HashSet`  should be O(1) :)

povilasb:
You're right. 
In addition, I would use `HashSet<PeerInfo>` in all places except for `self.peers`. The only reason `peers` is `Vec` because as Spandan mentioned we require to retain the insertion order. This way we have cached peers first and then hard coded ones - just a requirement we have :)

Also, if you extracted this peer list related code to a separate function and wrote unit tests, that would be awesome! :beer:

PR #1110, opened 2019-01-13
Filter out duplicate PeerInfos from bootstrap's peer list

1uka:
Since #1105 was closed, here is the update in a new PR.

I added a function called `update_peers` in the bootstrap module. Instead of calling `.contains()` on a `Vec`, I turn the peers list into a `HashSet` of references (once O(n) + time to form `HashSet` instead of n-times O(n), or O(n^2)). I added unit tests to make sure that the list doesn't contain duplicates and that after the peers list is shuffled again, the hard-coded PeerInfos stay at the end, after the cached ones. I did not change `Vec<PeerInfo>` to `HashSet<PeerInfo>` in other places such as [this](https://github.com/maidsafe/crust/blob/master/src/service_discovery/mod.rs#L34) because I think it would have affected performance.

I hope this is OK 🍻

### DIFF
--- a/src/main/bootstrap/mod.rs
+++ b/src/main/bootstrap/mod.rs
@@ -27,6 +27,7 @@ use socket_collection::TcpSock;
 use std::any::Any;
 use std::cell::RefCell;
 use std::collections::HashSet;
+use std::iter::FromIterator;
 use std::mem;
 use std::net::SocketAddr;
 use std::rc::{Rc, Weak};
@@ -246,10 +247,7 @@ impl<UID: Uid> State<BootstrapCache> for Bootstrap<UID> {
         }
 
         let rx = unwrap!(self.sd_meta.take()).rx;
-
-        while let Ok(listeners) = rx.try_recv() {
-            self.peers.extend(listeners);
-        }
+        receive_peers_into(&mut self.peers, rx);
 
         self.begin_bootstrap(core, poll);
     }
@@ -284,7 +282,7 @@ pub fn cache_peer_info(core: &mut EventLoopCore, peer_info: PeerInfo, config: &C
 }
 
 struct ServiceDiscMeta {
-    rx: Receiver<Vec<PeerInfo>>,
+    rx: Receiver<HashSet<PeerInfo>>,
     timeout: Timeout,
 }
 
@@ -294,7 +292,7 @@ fn seek_peers(
     core: &mut EventLoopCore,
     service_discovery_token: Token,
     token: Token,
-) -> crate::Res<(Receiver<Vec<PeerInfo>>, Timeout)> {
+) -> crate::Res<(Receiver<HashSet<PeerInfo>>, Timeout)> {
     if let Some(state) = core.get_state(service_discovery_token) {
         let mut state = state.borrow_mut();
         let state = unwrap!(state
@@ -334,6 +332,19 @@ fn shuffled_bootstrap_peers(
 
     peers.retain(|peer| !blacklist.contains(&peer.addr));
     peers
+}
+
+/// Appends peers received from `ServiceDiscMeta` to the peers list
+fn receive_peers_into(peers: &mut Vec<PeerInfo>, rx: Receiver<HashSet<PeerInfo>>) {
+    let our_peers: HashSet<&PeerInfo> = HashSet::from_iter(peers.iter());
+
+    let discovered_peers: HashSet<PeerInfo> = rx
+        .try_iter()
+        .flatten()
+        .filter(|peer| !our_peers.contains(peer))
+        .collect();
+
+    peers.extend(discovered_peers);
 }
 
 #[cfg(test)]
@@ -377,7 +388,7 @@ mod tests {
         }
     }
 
-    mod seek_pers {
+    mod seek_peers {
         use super::*;
 
         #[test]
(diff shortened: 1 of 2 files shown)


povilasb:
Cool, will take a look in a bit once I'm done with my current work ;)

src/main/bootstrap/mod.rs
```diff
@@ -430,6 +440,58 @@ mod tests {
...
+
+            let peers_len = peers.len();
+            let mut peers_set = HashSet::new();
+
+            assert!(peers.into_iter().all(|peer| peers_set.insert(peer)));
+            assert_eq!(peers_len, peers_set.len());
+        }
+
+        #[test]
+        fn it_preserves_order_if_shuffled_again() {
```
povilasb: When I read 
```
update_peers
    it_preserves_order_if_shuffled_again
```
It sounds like `update_peers()` has something to do with the peer shuffling... actually it does not.
I feel like the test would be a lot simpler, if we didn't include all that shuffle code. `update_peers()` simply appends peers from `rx: Receiver` to the given `Vec` and does not change the order of already inserted peers. Maybe that's what you could test? :)

src/main/bootstrap/mod.rs
```diff
@@ -332,6 +330,18 @@ fn shuffled_bootstrap_peers(
     peers
 }
 
+fn update_peers(peers: &mut Vec<PeerInfo>, rx: Receiver<Vec<PeerInfo>>) {
```
povilasb: The name is very generic, maybe `receive_peers_into()` or some docs about what the function does? :)

src/main/bootstrap/mod.rs
```diff
@@ -430,6 +440,58 @@ mod tests {
...
+
+            let (tx, rx) = mpsc::channel();
+
+            let duplicates: Vec<PeerInfo> = peers.iter().take(5).map(|peer| peer.clone()).collect();
+            tx.send(duplicates).unwrap();
+
+            update_peers(&mut peers, rx);
+
+            let peers_len = peers.len();
+            let mut peers_set = HashSet::new();
```
povilasb: What about simple `let peers_set: HashSet<PeerInfo> = peers.into_iter().cloned().collect();`?

src/main/bootstrap/mod.rs
```diff
@@ -332,6 +330,18 @@ fn shuffled_bootstrap_peers(
     peers
 }
 
+fn update_peers(peers: &mut Vec<PeerInfo>, rx: Receiver<Vec<PeerInfo>>) {
+    let discovered_peers: Vec<PeerInfo> = {
+        let our_peers: HashSet<&PeerInfo> = HashSet::from_iter(peers.iter());
+        rx.try_iter()
+            .flatten()
+            .filter(|peer| !our_peers.contains(peer))
```
povilasb: since `rx` receives `Vec<PeerInfo>`, theoretically speaking it could receive duplicate items itself. So say `rx.try_iter()` yields `["192.168.1.100", "192.168.1.100"]` (don't mind me skipping ports), then this list will be `flattened()` and `.filter()` receives each address individually. Note that at the time `filter()` gets second address, it is not in `our_peers` yet, so it will be passed through filter and later collected to `discovered_peers`.
I'd say at the bery least make `discovered_peers` a `HashSet` and see if you can make `rx` into `Receiver<HashSet<..>>`.

src/main/bootstrap/mod.rs
1uka: Sure, I will update it. At first I thought it would be important to see that even after appending to the list, if we shuffle it again we would have the wanted order. But since `peers` still is a `Vec<PeerInfo>` you are right, there is no need to test for that here, it is up to `shuffled_bootstrap_peers`.

src/main/bootstrap/mod.rs
1uka: I wanted to be able to assert if any of the `.insert()` calls would fail, which means the list contains duplicates, but I guess only checking if the length of the set and the vector are the same will be enough, so I will change that too.

1uka:
Ok, done. I made changes according to the things we discussed.

src/main/bootstrap/mod.rs
```diff
@@ -455,44 +456,36 @@ mod tests {
...
-            assert!(peers.into_iter().all(|peer| peers_set.insert(peer)));
-            assert_eq!(peers_len, peers_set.len());
+            assert_eq!(peers.len(), peers_set.len());
         }
 
         #[test]
-        fn it_preserves_order_if_shuffled_again() {
-            let hard_coded_peer = peer_info_with_rand_key(ipv4_addr(1, 2, 3, 4, 4000));
-            let cached_peers: HashSet<PeerInfo> = (0..10)
+        fn it_preserves_order_of_old_peers() {
```
povilasb: Could you add one more test to see if the peers from `rx` are actually added? Now we have tests to see if 
1. duplicates are omitted 
2. order is preserved
but there's no test that checks if peers from `rx` are added :/

src/main/bootstrap/mod.rs
```diff
@@ -336,13 +336,13 @@ fn shuffled_bootstrap_peers(
...
 /// Appends peers received from `ServiceDiscMeta` to the peers list
 fn receive_peers_into(peers: &mut Vec<PeerInfo>, rx: Receiver<HashSet<PeerInfo>>) {
-    let discovered_peers: HashSet<PeerInfo> = {
-        let our_peers: HashSet<&PeerInfo> = HashSet::from_iter(peers.iter());
-        rx.try_iter()
-            .flatten()
-            .filter(|peer| !our_peers.contains(peer))
-            .collect()
-    };
+    let our_peers: HashSet<&PeerInfo> = HashSet::from_iter(peers.iter());
```
povilasb: Do we really need to specify `our_peers` type?

1uka: If I omit the type, this is what happens:

```rust
error[E0282]: type annotations needed
   --> src/main/bootstrap/mod.rs:339:21
    |
339 |     let our_peers = HashSet::from_iter(peers.iter());
    |         ---------   ^^^^^^^^^^^^^^^^^^ cannot infer type for `S`
    |         |
    |         consider giving `our_peers` a type

```
So I guess there has to be a type annotation for `our_peers`

src/main/bootstrap/mod.rs
1uka: Sure, done.

1uka:
Hey, did I do it right, the rebasing and squashing? 😄

povilasb:
mhm, all good :+1:

PR #1111, opened 2019-01-15
resolves #1089: reduce the use of mutexes

### LINKED ISSUE #1089: Minimise Mutexes if possible
Go through all places we are currently using `Arc<Mutex<>>` pattern such as [here for e.g.,](https://github.com/maidsafe/crust/blob/bafde11bd33474ca3cfe2357d9b2d41364cdb831/src/main/service.rs#L49) and see if we can convert it to `Rc<RefCell<>>` by storing the data in the event loop thread somewhere (e.g. thread-local storage etc.) and accessing it only in that thread.

povilasb:
Makes code more simple and easier to reason about.


src/main/service.rs
```diff
@@ -87,6 +87,9 @@ pub struct Service<UID: Uid> {
     our_uid: UID,
     our_pk: PublicEncryptKey,
     our_sk: SecretEncryptKey,
+    /// Used to send our listener addresses to service discovery server.
+    service_discovery_listeners_tx: mpsc::Sender<HashSet<PeerInfo>>,
+    service_discovery_listeners_rx: Option<mpsc::Receiver<HashSet<PeerInfo>>>,
```
ustulation: Hmm it looks like the pattern should be reversed here. Keeping an optional in `Service` etc., seems a bit other-way-round. It's the `ConnListener` to whom the observers are subscribing to and `ServDisc` happens to be one of the observers. So,

### 1. A full fledged pattern
`ConnListener` keeps something like:
```rust
listener_addr_observers: Vec<Sender<SocketAddr>>,
```
and try to:
```rust
self.listener_addr_observers.retain(|obs| obs.send(blah).is_ok());
```
Subsequently, during `ServDisc` creation, do create a  channel, give rx to `ServDisc` and `push` the tx to `ConnListener`.
Given that there's only one ever obs in this case the "full-fledgedness" could be an overkill (just maybe, not by much though). So an alternative could be:

### 2. API in `ServDisc` to be called explicitly
Each time you get a new addr for direct listeners, get the `State` of `ServDisc` as usual from `Core` and invoke an API in it to add the new listener addr.

### 3. Use the persistent state (user-data) stored in the event loop thread's parent stack.
Let `ServDisc` ask for something that impls a trait it defines:
```rust
// in serv-disc module:
trait GetDirectListeners {
    fn direct_listeners(&self) -> HashSet<SocketAddr>;
}

ServiceDiscovery<T: GetDirectListeners> {...}
```
then, wherever you have `CrustUserData` (or whatever it's being currently called):
```rust
impl GetDirectListeners for CrustUserData { ... }
```

IMO:
1. is better if you had constantly chaning listener addrs and many observers, both of which we don't. So looks a bit forced (just like what's there currently with `Service` holding and optional etc., but i would choose this one over that)
2. will need a downcast_mut so not very design friendly although you get the most bang for your buck (short and not-so-clean solution)
3. is the more expressive (of the intent) and clear design and hence i think it's the best of all.

povilasb:
We went with the 3rd option: store our global addresses in `Core`, make `CrustData` implement specific trait that is used by `ServiceDiscovery` as well.

PR #1115, opened 2019-01-18
Limit bootstrap peers

### LINKED ISSUE #1106: Limit bootstrap peers
Currently Crust does 2 things we don't like:

1. bootstrap cache has no limit. That can cause the system to behave unpredictibly: consume too much memory in RAM and/or disk, act slow where it's not expected, like loading the cache, etc.
2. `Service::bootstrap()` will attempt all peers in parallel, even if there are thousands of cached peers. The problem with this is that many systems won't allow you to use more than ~1000 sockets in parallel.

So 

1. introduce configurable bootstrap cache limit with the reasonable defaults (say 10000 which roughly results in 40 kB cache). When bootstrap cache exceeds the limit, remove the contacts with smallest trust level - see bellow for description.
2. Introduce a counter for bootstrap cache entries that logs how many times a peer was used to successfully bootstrap off. Then this counter will be used as a trust metric - the more we've bootstrap from the peer the more we trust it's availability.
3. Limit parallel connections during bootstrap to prevent from exhausting system sockets.

1uka:
I tackled point 1 and 2 from #1106 in this PR.

I added configurable bootstrap cache limit, implement trust metric for the peers by changing the container type to `HashMap<PeerInfo, Instant>` so that the most recently active peers are always kept in the cache. I used `time::Instant` because of the reasons I mentioned [in this comment](https://github.com/maidsafe/crust/issues/1106#issuecomment-455526538). Sorry I did not wait for your feedback, but this was the time I had allocated to work on this issue for today.

Partially resolves #1106

### DIFF
--- a/src/main/bootstrap/cache.rs
+++ b/src/main/bootstrap/cache.rs
@@ -10,9 +10,12 @@
 use crate::common::PeerInfo;
 use config_file_handler::{self, FileHandler};
 use std::cell::RefCell;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::ffi::OsString;
 use std::rc::Rc;
+use std::time::Instant;
+
+const BOOTSTRAP_CACHE_DEFAULT_LIMIT: usize = 1500;
 
 /// Reference-counted bootstrap cache - keeps log of known publicly accessible peers.
 #[derive(Clone)]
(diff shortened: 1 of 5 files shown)


povilasb:
Note that we've just pushed a lot of changes to the master branch, so you better rebase you work of it :) Those changes remove `Uid` trait so it touches most parts of the code, but shouldn't conflict with what you're doing :)

1uka:
Thanks, any other way I can help and contribute to this? Or any other issue in particular.

povilasb:
@1uka thanks for the interest, I'll see what beginner friendly issues we have. Maybe you have some preferences you'd be interested in yourself? :)

Anyway, closing in favor of https://github.com/maidsafe/crust/pull/1120

povilasb:
Mhm, thanks for the contributions. I know, everybody's busy :) We are currently going through the Crust roadmap to see what to prioritize. Once done, will see what low hanging fruits we will have.

Btw, there's a formal algorithm (https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU)) very similar to what you've described with `VecDeque` :).  Anyway, good thinking :+1:

PR #1119, opened 2019-01-24
resolves #1118: accept encryption keys via Service constructor

### LINKED ISSUE #1118: Pass node public and secret keys to Service constructor
Currently Crust generates it's own set of public/secret encryption keys. Better pass from the upper layer. This has several advantages:

1. Node key can be pregenerated and reused multiple times it's started
2. The same encryption keys can be reused from the upper layers

povilasb:
This allows crust to reuse encryption keys from the upper layers.  Since
peer ID holds public key as well, a lot of messages, structures were
holding redundant copy of peer public key. This commit removes that as
well.

### DIFF
--- a/src/tests/utils.rs
+++ b/src/tests/utils.rs
@@ -15,7 +15,7 @@ use maidsafe_utilities::event_sender::{MaidSafeEventCategory, MaidSafeObserver};
 use mio_extras::channel::channel;
 use mio_extras::timer;
 use rand;
-use safe_crypto::{gen_encrypt_keypair, gen_sign_keypair};
+use safe_crypto::{gen_encrypt_keypair, gen_sign_keypair, SecretEncryptKey};
 use std::env;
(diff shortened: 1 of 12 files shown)


src/tests/utils.rs
```diff
@@ -42,13 +42,14 @@ macro_rules! expect_event {
 }
 
 /// Generates random peer id.
-pub fn rand_peer_id() -> PeerId {
-    let (enc_pk, _enc_sk) = gen_encrypt_keypair();
+pub fn rand_peer_id() -> (PeerId, SecretEncryptKey) {
```
ustulation: `rand_peer_id` -> `rand_peer_id_and_enc_sk` as it was confusing to what sec-key was being returned in above (as you did not qualify the vars saying `peer_enc_sk` or `peer_sign_sk` but merely `peer_sk`. So if the function is appropriately named, it'll be less confusing.

PR #1120, opened 2019-01-29
resolves #1106: bootstrap cache improvements

### LINKED ISSUE #1106: Limit bootstrap peers

povilasb:


src/main/bootstrap/cache.rs
```diff
@@ -48,53 +47,66 @@ impl Cache {
...
-        let inner = self.inner.borrow();
-        file_handler.write_file(&inner.peers)?;
+        file_handler.write_file(&self.ordered_peers)?;
         Ok(())
     }
 
-    /// Returns current snapshot of peers in the cache.
-    pub fn peers(&self) -> HashSet<PeerInfo> {
-        self.inner.borrow().peers.clone()
+    /// Returns given amount of cached peers and updates their last used time in the cache, hence
```
ustulation: Comment about time no longer valid

src/main/bootstrap/mod.rs
```diff
@@ -39,6 +39,9 @@ const SERVICE_DISCOVERY_TIMEOUT_SEC: u64 = 1;
 const BOOTSTRAP_TIMER_ID: u8 = 0;
 const SERVICE_DISCOVERY_TIMER_ID: u8 = BOOTSTRAP_TIMER_ID + 1;
 const MAX_CONTACTS_EXPECTED: usize = 1500;
+/// Bootstrap will use up to max this amount of cached peers. We don't want to exhaust socket
```
ustulation: Neither this comment nor the impl is correct. We do want to go through the whole collection of everyone in the cache and in hardcoded contact list and via serv-disc. Just that we want to do max 10 at at time in parallel. It was always about parallel bootstrap limit.

So the last point [here](https://github.com/maidsafe/crust/issues/1106) is:
> Limit parallel connections during bootstrap to prevent from exhausting system sockets.

So the comment should indicate that we try this number of connection attempts in parallel at any given time. When there's a failure, take the next one up from the list and so on until the whole combined list is exhausted.

src/main/bootstrap/cache_validator.rs
```diff
@@ -0,0 +1,198 @@
...
+            self.ping_peer_and_cache_alive(core, poll, peer);
+        }
+    }
+
+    /// Send STUN request to given peer and wait for response.
+    /// If response arrives within timeout, peer is added to the bootstrap cache head.
+    fn ping_peer_and_cache_alive(&mut self, core: &mut EventLoopCore, poll: &Poll, peer: PeerInfo) {
+        let self_weak = self.self_weak.clone();
+        let expired_tx = self.expired_tx.clone();
+        let on_done = move |core: &mut EventLoopCore,
```
ustulation: These were all called `finish` (and it wasn't me who did that, it probably belonged to 2016 naming convention set by Andrew). Why change all that ? Pls revert back to `finish` and report discrepancies if you find any as that should also be corrected. I don't remember `on_done` anywhere in Crust before so pls refrain from changing those.

src/main/bootstrap/cache_validator.rs
```diff
@@ -0,0 +1,198 @@
...
+impl State<CrustData> for CacheValidator {
+    fn ready(&mut self, core: &mut EventLoopCore, poll: &Poll, kind: Ready) {
+        if kind.is_readable() {
+            self.read_expired_rx(core, poll)
+        }
+    }
+
+    fn terminate(&mut self, core: &mut EventLoopCore, poll: &Poll) {
+        self.terminate_requests(core, poll);
+        let _ = core.remove_state(self.token);
```
ustulation: missing de-registeration of the `expired_rx` which remains in the poll-list as an orphan (i.e., without any associate `State` with it).

src/main/bootstrap/cache_validator.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// Starts running bootstrap cache entry validation.
+    pub fn start(
+        core: &mut EventLoopCore,
+        poll: &Poll,
+        our_pk: PublicEncryptKey,
+        our_sk: SecretEncryptKey,
+    ) -> crate::Res<()> {
+        let token = core.get_new_token();
+
+        let (expired_tx, expired_rx) = channel();
```
ustulation: I'm not liking this mio::channel to communicate between states all residing intra-thread. Channels are more for cross threaded communication so I'm thinking we can have a different design but need to jump somewhere else just now - will come back to this. Basically it feels like an overkill - like using Arc's and Mutexes and Atomics for intra-thread communication with needless overhead. However this might be the best compromise but will need to come back to this to see if there's no alternate/better way.

povilasb: Mio async channel was used for notifications. Under the hood it should be just an in-mem buffer with some notification mechanism. So it shouldn't introduce more overhead compare to other solutions, cause some sort of buffering will be used anyway.

src/main/bootstrap/cache_validator.rs
povilasb: `finish` is an imperative form and `on_finish`/`on_done` sounds better to me as that implies callback. Anyway this is a minor thing and I'll update it. But having something settled is not the best argument to avoid refactoring to me :)

ustulation: > But having something settled is not the best argument to avoid refactoring to me :)

Sounds good for an oxymoron :P

src/main/bootstrap/cache_validator.rs
ustulation: What about instead of having a sender to the cache-validator kept as user-data in `Core` and using that to go via Poll indirectly, have `CacheValidator` itself (as an `Rc<RefCell` which is what it is here anyway) lodged there ? So now we can have direct invocations of `CacheValidator` facilities which prevents going through a channel route. This is certainly less overhead than if we used channels is it not ? Considering the channels would have to do additional work including thread safety guarantees (when we don't need it) and posting to the Poll loop.

src/main/bootstrap/cache_validator.rs
```diff
@@ -149,6 +125,15 @@ impl State<CrustData> for CacheValidator {
     }
 }
 
+/// Schedules tests for inactive cached bootstrap peers.
+pub fn test_inactive_cached_peers(core: &mut EventLoopCore, poll: &Poll, peers: HashSet<PeerInfo>) {
+    if let Some(state) = core.get_state(EventToken::BootstrapCacheValidator.into()) {
+        if let Some(validator) = state.borrow_mut().as_any().downcast_mut::<CacheValidator>() {
```
ustulation: It is to avoid this downcast i was [mentioning](https://github.com/maidsafe/crust/pull/1120#discussion_r256052331) about having it in the Core's userdata directly. Would that be a better pattern ? Because as we can see there's not really anything from the `State` trait that this is using so why even make it a `State` was my reasoning.

povilasb: I actually experimented with this in the morning and the problem was that I had to terminate the `GetExtAddr` request states. I thought of doing that during `drop()` of `CacheValidator` but I can't access `poll` and `core`, hence this implementation.

Any ideas how I could nicely terminate the [sent_requests](https://github.com/maidsafe/crust/pull/1120/files#diff-7e42e856fb45c217a3ce4b3d9f24650bR33) without `CacheValidator` being a `State`?

ustulation: Can it not have a non-state terminate member function just in the `impl CacheValidator` ?

povilasb: It can, but then there's no one preventing us from leaking those states except explicitly documenting that such destructor MUST be called.

PR #1123, opened 2019-02-15
fix/example: remove external reachability requirement

povilasb:
By default Crust requires nodes to be externally reachable. Although,
that can be disabled.  There's no point of this requirement in the
test, hence disabled it.


douglascaetano:
Fixed on PR #1141.

PR #1124, opened 2019-02-18
closes #835: linting

### LINKED ISSUE #835: Lint
Bring back the `bad_style` lint (previously in `forbid` section and needed to be removed due to serde) as `deny` and specifially `allow` the `non-camel-case` required by serde [as in S.E for e.g.](https://github.com/maidsafe/self_encryption/blob/4857d48e05b36a6af506008369f7962720b5a43f/src/lib.rs#L145)

povilasb:
Deny `nonstandard-style`: https://doc.rust-lang.org/rustc/lints/groups.html

### DIFF
--- a/examples/crust_peer.rs
+++ b/examples/crust_peer.rs
@@ -300,6 +300,7 @@ fn main() {
     let (peer_id, peer_sk) = new_peer_id();
     let mut service = unwrap!(Service::with_config(event_sender, config, peer_id, peer_sk));
     unwrap!(service.start_listening_tcp());
+    unwrap!(service.set_ext_reachability_test(false));
     service.start_service_discovery();
     let service = Arc::new(Mutex::new(service));
     let service_cloned = service.clone();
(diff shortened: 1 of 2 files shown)


examples/crust_peer.rs
```diff
@@ -300,6 +300,7 @@ fn main() {
     let (peer_id, peer_sk) = new_peer_id();
     let mut service = unwrap!(Service::with_config(event_sender, config, peer_id, peer_sk));
     unwrap!(service.start_listening_tcp());
+    unwrap!(service.set_ext_reachability_test(false));
```
ustulation: Ah OK if it were a test. Since this is an example should we put a sample-config and set it there and allow users to set it accordingly too ? Some might want to use this in different networks where ext-reach. might be wanted by them (even to test).

povilasb:
Incorporated these changes into https://github.com/maidsafe/crust/pull/1132

PR #1132, opened 2019-02-22
closes #835: update bootstrap cache on msg send

### LINKED ISSUE #835: Lint

povilasb:
Move peer to the top of the bootstrap cache, if we are sending a message to it. That means this peer is the most active.

### DIFF
--- a/src/main/active_connection.rs
+++ b/src/main/active_connection.rs
@@ -100,30 +100,31 @@ impl ActiveConnection {
     }
 
     fn read(&mut self, core: &mut EventLoopCore, poll: &Poll) {
+        let mut msg_received = false;
         loop {
             match self.socket.read::<Message>() {
                 Ok(Some(Message::Data(data))) => {
                     let _ =
                         self.event_tx
                             .send(Event::NewMessage(self.their_id, self.their_role, data));
-                    self.reset_receive_heartbeat(core, poll);
-                    self.updated_bootstrap_cache(core, poll);
+                    msg_received = true;
                 }
-                Ok(Some(Message::Heartbeat)) => {
-                    self.reset_receive_heartbeat(core, poll);
-                    self.updated_bootstrap_cache(core, poll);
-                }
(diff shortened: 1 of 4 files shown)


src/main/active_connection.rs
```diff
@@ -96,13 +107,16 @@ impl ActiveConnection {
...
+                    self.updated_bootstrap_cache(core, poll);
                 }
                 Ok(Some(Message::Heartbeat)) => {
                     self.reset_receive_heartbeat(core, poll);
+                    self.updated_bootstrap_cache(core, poll);
                 }
                 Ok(Some(message)) => {
                     debug!("{:?} - Unexpected message: {:?}", self.our_id, message);
                     self.reset_receive_heartbeat(core, poll);
+                    self.updated_bootstrap_cache(core, poll);
```
ustulation: why are we updating it in a loop ? Wouldn't it be better to do it for `Ok(None)` case only (so opposite of what you've done here) ? That way most of the time it's triggered only _after_ all the reading has been done for this time from the peer (instead of after every call of the sock-coll read which might be returning things from buffer).

If we are very pernickety, then we might say there might be spurious wake ups and read called without there being anything to actually read which will also return `Ok(None)` . For that either live with it OR have some sort of a bool before the loop and toggle it once anything legit is read and then check that in `Ok(None)` and update the BStrap cache there. Doing it in a loop seems wasteful other wise (and unfair too in a way as theoretically all the msgs could have been read and buffered by the sock-coll internally which does not mean that peer is still sending when we are reading the next time in the loop - so timestamp shouldn't be updated)

povilasb:
Deprecated in favor of https://github.com/maidsafe/crust/issues/1138

PR #1137, opened 2019-02-26
fix/example: load config from file

douglascaetano:
Reads config from file instead of generating it from the default values.


examples/crust_peer.rs
```diff
@@ -283,10 +283,7 @@ fn main() {
         crust_event_category,
         category_tx,
     );
-    // TODO {{{
-    // let mut config = unwrap!(::crust::read_config_file());
-    let mut config = Config::default();
-    // }}}
+    let mut config = unwrap!(read_config_file());
```
ustulation: you have not provided any config file - what is it supposed to read ? Does this not panic ?

douglascaetano: This is the API from Crust itself. This function is defined in `main/config_handler.rs` and uses a predefined name for the config file. I presumed that this was the correct way to read the file because of the previous comment in-place.

As of the panic, I must admit I hadn't thought about it, as the code was already structured like in my commit on the previous comment. I went through the code now and the `FileHandler` struct used in `read_config_file()` creates the file if it doesn't exist based on the struct type of the requested data (in this case, `Config`). It would panic if can't create the new file on one of the following paths: binary directory, user home or system cache directory.

ustulation: > the `FileHandler` struct used in `read_config_file()` creates the file if it doesn't exist

Ah that's all i was looking for - so it's not going to panic so that's good.

Thinking again on this point however, i wasn't meaning to specify a file name but to place a sample config next to the example so that the user can tweak it if necessary. In that keep the ext-reachability as false by default. Otherwise there needs to be too much knowledge to do that i.e., user has to know that the config will be created in the first run, then obtain it that way to tweak it. If it's simply placed next to the example it makes life easier for the person trying out the example.

Also can we pls change the name to chat example or something more attractive ?

@douglascaetano

S-Coyle:
Reviewed this from a QA perspective @douglascaetano and the max_size works perfectly 😄 
With the change from TLRU to LRU I assume the timeout will shortly no longer be required?

examples/crust_peer.rs
douglascaetano: > place a sample config next to the example

How should we do it? I mean, we can add a sample config on the repo, but the build process won't copy it to the binary folder automatically. Is it okay?

Anyway, I'm kind of against it because we would need to remember to update the config file whenever we change the `Config` struct. (Unless we used the pre-build hook from Cargo, about which I read in [the Cargo Book](https://doc.rust-lang.org/cargo/reference/build-scripts.html) this week, to write a new config file based on the current struct whenever we build the binary.)


> user has to know that the config will be created in the first run

An alternative would be to add a parameter like e.g. `--generate-config` in which the binary would only create a new config file (if it doesn't exist yet) and exit. Then the user could tweak it as it sees fit.


> Also can we pls change the name to chat example or something more attractive ?

Yes, we can! :+1: Would `simple_chat` fit? Or `simple_connection` maybe, as it is not exactly a chat app.

ustulation: >  I mean, we can add a sample config on the repo, but the build process won't copy it to the binary folder automatically. Is it okay?
If you run it via `cargo run --example <blah>` does it pick from next to the example source code ? If yes then that's OK otherwise i agree, it needs to be put next to the binary by the user.

> Unless we used the pre-build hook from Cargo, about which I read in [the Cargo Book](https://doc.rust-lang.org/cargo/reference/build-scripts.html) this week, to write a new config file based on the current struct whenever we build the binary.

If available i would be for this. This is an example, something like an app that uses crust. So it should demo that there's this config which can be tweaked etc. `--generate-config` looks like an overkill for a simple demo of an example which some might just want to use it in the minimalistic way. However i agree it's _a_ way (and maybe more full fledged).

These sample-configs are also in other maidsafe repos (like in maidsafe_utilities etc.,) alongside examples or in a separate directory.

> because we would need to remember to update the config file
Yes, same as you would have to remember to update the example because just building (which the CI will assist you in doing) is not a complete picture for an executable - you need to look at the runtime behaviour too and hence the need to re-visit on code changes to see if it's still doing what you expect. The CI only helps in running the test executables so that's fine; for examples you have some manual-dev work overhead anyway - they don't semver utilise your library - they just use whatever is the recent code in the `src/` folder.

ustulation: >  Or `simple_connection` maybe, as it is not exactly a chat app.
Let's call it `secure-p2p-connection` then ? Just to make sure we don't undersell what it's demoing - else ppl would wonder if it's just a connector connecting to an acceptor, which is like a few lines of code.

douglascaetano: > > I mean, we can add a sample config on the repo, but the build process won't copy it to the binary folder automatically. Is it okay?
> 
> If you run it via `cargo run --example <blah>` does it pick from next to the example source code ? If yes then that's OK otherwise i agree, it needs to be put next to the binary by the user.

No, the function that loads the config file only look for the directories I listed before: binary directory, user home or system cache directory. So it would not suffice to keep the file together with the source.

> > Unless we used the pre-build hook from Cargo, [...]
> 
> If available i would be for this.

Thinking again, this may be even more overkill than the `--generate-config` parameter, at least for now, because it would take a lot more time to develop, as I just discovered about this and don't know how to properly use.

But I do like the idea of generating an example config file that way. We could discuss it in another moment.

> These sample-configs are also in other maidsafe repos (like in maidsafe_utilities etc.,) alongside examples or in a separate directory.
>
> [...]
>
> > because we would need to remember to update the config file
> 
> Yes, same as you would have to remember to update the example [...]

Well, yes, I think. I'll go this way, then.

ustulation:
Merged in https://github.com/maidsafe/crust/pull/1141

PR #1139, opened 2019-02-28
closes #1138: LRU bootstrap cache

### LINKED ISSUE #1138: Simplify bootstrap cache
## The problem

Currently Crust will evict peers from bootstrap cache, if they are being unresponsive for too long (2 minutes). The problem with this approach is that peers might become active in some time, e.g. a minute if he's only rebooting the machine. So we will remove peer from cache that potentially could help us bootstrap to the p2p network in the future. In addition, this happens even if the cache has not reached its limit.

## Solution

1. If the cache is not full and previously cached peer is not responsive, don't evict it from the cache.
2. Evict peer from the bootstrap cache only when cache is full (has 200 entries) and we find out about a new directly accessible peer which we'd like to cache.
3. So remove all the time based external reachability tests of cached peers.

povilasb:
Evict peers from bootstrap cache only when the cache is full and we want to cache a new peer.


src/main/bootstrap/cache.rs
```diff
@@ -82,25 +74,15 @@ impl Cache {
...
-    /// ## Returns
-    ///
-    /// A list of expired peers.
-    pub fn put(&mut self, peer: PeerInfo) -> HashSet<PeerInfo> {
-        let (_, mut expired) = self.peers.notify_insert(peer.pub_key, peer);
-        expired.drain(..).map(|(_key, val)| val).collect()
-    }
-
-    /// Removes given peer from the cache.
-    pub fn remove(&mut self, peer: &PeerInfo) {
```
ustulation: I think we should still have the remove API even if we don't use it just now. I am thinking about cases where we _know_ that a peer is malicious. In that case i don't think it makes sense to keep them in the cache (if we are blacklisting them or something) even if it mean decrementing the peer-count in cache.

src/main/active_connection.rs
```diff
@@ -412,7 +412,9 @@ mod tests {
                 let state = unwrap!(core.get_state(sock_token));
                 let mut state = state.borrow_mut();
                 let active_conn = unwrap!(state.as_any().downcast_mut::<ActiveConnection>());
-                wait_for(sock_token, &poll);
```
ustulation: I don't see why this should fire at all in a deterministic way. You have not registered this socket (`peer1_sock`) with poll to hear about readable readiness so even though the listener writes something, there won't be an notification fired.

src/main/active_connection.rs
```diff
@@ -267,3 +287,150 @@ enum HeartbeatAction {
...
+                    CrustUser::Client,
+                    event,
+                    event_tx,
+                );
+
+                let state = unwrap!(core.get_state(sock_token));
+                let mut state = state.borrow_mut();
+                let active_conn = unwrap!(state.as_any().downcast_mut::<ActiveConnection>());
+                // TODO(povilas): debug why wait_for blocks indefinitely on Mac OS
+                //wait_for(sock_token, &poll);
```
ustulation: I don't see why this should fire at all in a deterministic way. You have not registered this socket (`peer1_sock`) with poll to hear about readable readiness so even though the listener writes something, there won't be an notification fired.

povilasb: It's being registered at https://github.com/maidsafe/crust/pull/1139/files#diff-bf742b5e6a9703a78753d7e485151c7dR386

ustulation: @povilasb i don't think i got you. What i said was:
> You have not registered this socket (`peer1_sock`) with poll to hear about **readable** readiness

What you point to has:
```
unwrap!(poll.register(&peer1_sock, sock_token, Ready::writable(), PollOpt::edge()));
```

Can you explain more how is that going to generate a notification if the listener sent something ?

douglascaetano: I found the problem: it is in the `ActiveConnection::start()` function. This function calls `read()` in its end, before returning, effectively consuming all data in socket's receiving buffer. That means that no `readable` event will ever be raised in this test. It passed previously because of some still unknown reason for `writable` event to be raised again, even though the event was registered as edge-triggered (and was already triggered before).

PR #1141, opened 2019-03-01
fix/example: disables external reachability by default

douglascaetano:
The external reachability test was making connections fail when the example was run in a local network, so disables the test by default.

A flag `--test-extreach` was added to enable the test whenever needed.

The previous PR #1123 from Povilas had a race condition in which the test could be disabled before the listener was ready.


examples/secure_p2p_connection.rs
```diff
@@ -393,6 +402,16 @@ fn main() {
                                     peer_id,
                                 );
                             }
+                            crust::Event::ListenerStarted(port) => {
+                                println!("\nListener started on port {}", port);
+                                if !test_extreach {
+                                    println!("Test for External Reachability is disabled");
+                                    let service_unlocked = unwrap!(service.lock());
```
ustulation: Reads strange that you are getting something called unlocked by calling something called `lock()` - maybe call it service_guard or something ?

douglascaetano: Hmmm, yeah, that sounds better. I'll change it.

PR #1155, opened 2019-05-28
Version change to 0.32.0

lionel-faber:
- Update Travis script and maidsafe_utilities dependency 
- Version change to 0.32.0


.travis.yml
```diff
@@ -10,9 +10,6 @@ language: rust
 rust:
   - 1.29.2
```
nbaksalyar: Needs to be replaced with `stable`:
```suggestion
  - stable
```

nbaksalyar (changes requested):

mrcnski (changes requested):
Looks like this PR is missing some clippy fixes:
https://github.com/maidsafe/crust/compare/legacy...m-cat:legacy-clippy

nbaksalyar (dismissed):
Thanks @lionel1704 and @m-cat, good work! :+1:

